PR #47, opened 2018-10-30
Adding a new param wsservpath to shell html

notthetup:
This optional query param will be used to decide where to look for the shell websocket server.

So for example just requesting for 

```html
<iframe src="/fjage/shell/index.html" style="width: 1024; height: 768; padding: 5; background: black;"></iframe>
```

will have the shell javascript trying to connect to `<host:port>/shell/ws/`.

But 

```html
<iframe src="/fjage/shell/index.html?wsservpath=/fjage/shell/ws/" style="width: 1024; height: 768; padding: 5; background: black;"></iframe>
```

will have the shell javascript trying to connect to `<host:port>/fjage/shell/ws/`.

### DIFF
--- a/src/main/resources/org/arl/fjage/web/shell/index.html
+++ b/src/main/resources/org/arl/fjage/web/shell/index.html
@@ -32,7 +32,9 @@
   function setupHterm() {
     const term = new hterm.Terminal();
     term.onTerminalReady = function() {
-      const ws = new WebSocket('ws://'+window.location.hostname+':'+window.location.port+'/shell/ws/');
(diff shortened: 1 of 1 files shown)


mchitre:
I believe this shouldn't be required anymore, as I had fixed this in 38d4a66a51a653956ab3948382686318f5e71323 by getting path from the URL.

@notthetup See if that addresses the issue that you were trying to fix?

notthetup:
Yup. I guess that's why @AnneThKu didn't see this issue.

PR #52, opened 2018-11-14
adding DefaultHandler as per Jetty docs

notthetup:
Jetty documents suggest having a `DefaultHandler ` at last the last `Handler` after every Collection of Handlers so that errors are caught and replied with a good 404 response.

> Notice that a HandlerList is used with the ResourceHandler and a DefaultHandler, so that the DefaultHandler generates a good 404 response for any requests that do not match a static resource.

https://www.eclipse.org/jetty/documentation/current/embedding-jetty.html


mchitre:
Merged

PR #53, opened 2018-12-11
fix (connectors):Increasing Max Message size for Websockets

AnneThKu:


mchitre:
Merged into a separate branch, with intent to adopt changes into 1.5 with more flexibility to set buffer sizes on instantiation of the connector

PR #75, opened 2019-06-10
feat(fjage): add support for supplying an optional ip to bind to, and…

notthetup:
… default to binding to localhost only

The default `WebServer` constructor now will only bind to `localhost`. And a new constructor is available where the IP can be set to `0.0.0.0` to enable any device to connect.


mchitre:
Looks great!

PR #99, opened 2019-11-14
fix(fjagejs): proper encoding of JSON messages with non-ASCII chars.

ngyewch:


mchitre:
Could you PR to `dev` please? We merge to `master` only during releases. Also, commit message for `fix(remote)`, since the fix is applicable for not just JS, but all remote clients. Thanks.

ngyewch:
OK. Will redo this.

PR #103, opened 2019-11-21
fix(fjagejs): aligning to fjage behaviour by removing rejecting in ca…

notthetup:
…ses of timeout errors and resolving with a null object instead

### DIFF
--- a/src/main/resources/org/arl/fjage/web/fjage.js
+++ b/src/main/resources/org/arl/fjage/web/fjage.js
@@ -427,7 +427,7 @@ export class Gateway {
   // returns a Promise
   _websockTxRx(rq) {
     rq.id = _guid(8);
-    return new Promise((resolve, reject) => {
+    return new Promise(resolve => {
       let timer = setTimeout(() => {
         delete this.pending[rq.id];
         if (this.debug) console.log('Receive Timeout : ' + rq);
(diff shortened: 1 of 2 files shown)


mchitre:
Sounds fine to me

mchitre:
What is the impact on all web pages that currently use fjage.js?

notthetup:
It shouldn't be much. Mainly the change would have to be updating `try..catch` blocks to null checks. But I believe we already have null checks for the values returning by Promises. Will verify.

notthetup:
I take that back, we will need some changes to the WebUI. There is a lot of code of the following format.

```
gw.request(req).then((rsp) => {
      nodeAddr.innerHTML = rsp.value;
    }).catch((ex) => {
      console.warn('Couldn\'t get Node Address' + ex);
      nodeAddr.innerHTML = '';
    });
```

It will have to be changed to 

```
gw.request(req).then( rsp => {
  if (!rsp) console.warn('Couldn\'t get Node Address' + ex);
  nodeAddr.innerHTML = rsp ? rsp.value : '';
})
```

mchitre:
Better to put in 1.7 then, since it breaks backward compatibility?

mchitre:
Ready to merge?

notthetup:
Yup!

PR #116, opened 2020-01-30
fix(shell):fix to overwrite file contents in PutFileReq

AnneThKu:
When some content is removed from a file, the `RandomAccessFile` was not saving the file correctly with the modified content. Adding `setLength(ofs)`, where ofs value is 0, seems to fix the issue.

### DIFF
--- a/src/test/groovy/org/arl/fjage/test/spec/fjageSpec.js
+++ b/src/test/groovy/org/arl/fjage/test/spec/fjageSpec.js
@@ -3,6 +3,7 @@ import { Performative, AgentID, Message, Gateway, MessageClass } from '../../fja
 const DIRNAME = '/tmp';
 const FILENAME = 'fjage-test.txt';
 const TEST_STRING = 'this is a test';
+const NEW_STRING = 'new test';
 const GetFileReq = MessageClass('org.arl.fjage.shell.GetFileReq');
 const GetFileRsp = MessageClass('org.arl.fjage.shell.GetFileRsp');
 const ShellExecReq = MessageClass('org.arl.fjage.shell.ShellExecReq');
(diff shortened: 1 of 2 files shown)


notthetup:
I'm guessing the problem in the logic here is for truncation. If we're just truncating the file, the `ofs` and `content` aren't enough to define what to do with the rest of the file.

This one needs to be thought through a little more.

mchitre:
Its rare to need to write in the middle of the file. So I'd propose always truncating the file at the end of the written content, no matter from what offset it was written.

notthetup:
OK. In that case, we can do something like

`((RandomAccessFile) os).setLength(ofs+contents.length);` should suffice.

But we should also document this behavior somewhere, and ensure that we test all the intended behaviours.

https://github.com/org-arl/fjage/blob/dev/src/test/groovy/org/arl/fjage/test/spec/fjageSpec.js#L395

AnneThKu:
Added the change `((RandomAccessFile) os).setLength(ofs+contents.length);` and also added the test cases for `PutFileReq` in `fjageSpec.js`.

src/test/groovy/org/arl/fjage/test/spec/fjageSpec.js
```diff
@@ -414,6 +415,117 @@ describe('Shell GetFile/PutFile', function () {
       });
     });
   });
+
+  it('should be able to append a file using PutFileReq', function (done) {
```
notthetup: If you don't provide an `offset`, then it'll overwrite the file right? So is that considered appending? Or overwriting?

src/test/groovy/org/arl/fjage/test/spec/fjageSpec.js
```diff
@@ -414,6 +415,117 @@ describe('Shell GetFile/PutFile', function () {
...
+        expect(new TextDecoder('utf-8').decode(new Uint8Array(msg.contents))).toEqual(TEST_STRING + ' ' + TEST_STRING);
+        done();
+      }).catch((ex) => {
+        console.error(ex);
+        fail();
+      });
+    });
+  });
+
+  it('should be able to save the file using PutFileReq when some content is removed', function (done) {
```
notthetup: "should be able to correctly update the file using PutFileReq when some content is removed from the file"

src/test/groovy/org/arl/fjage/test/spec/fjageSpec.js
AnneThKu: Yes. it will be overwriting. I will update the description.

src/test/groovy/org/arl/fjage/test/spec/fjageSpec.js
AnneThKu: updated the description for all cases where offset is 0, <0 and >0.

PR #118, opened 2020-02-11
fjage.js: defensive check for null in _decodeBase64()

ngyewch:
# Sample error

Location: `fjage.js` `_decodeBase64()`
```
JSON Parsing error: TypeError: right-hand side of 'in' should be an object, got null
JSON : {"id":"a6145a373f06d7b85783585800eb66e0","inResponseTo":"agentForService","agentID":"remote","message":null}
```

# Probable cause
`typeof null` evaluates to `'object'`.


notthetup:
@AnneThKu You saw this as well right?

PR #122, opened 2020-02-28
Fjagejs inheritance

notthetup:
Support for inheritance in fjagejs MessageClass using an optional second parameter in the function. For eg. `MessageClass(classname, parentname)`

The diff looks huge, but most of the changes are because of reorganizing fjage.js to move all the exports to the top of the file.

Also added a `Services` class which exports the SHELL Service.


notthetup:
Tests seem to be failing somewhere else though.

```
org.arl.fjage.test.BasicTests > testShell FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertTrue(Assert.java:52)
        at org.arl.fjage.test.BasicTests.testShell(BasicTests.java:339)
```

PR #133, opened 2020-03-28
Parameter support in fjage.js

notthetup:
Adds support for parameter set/get in the fjage.js Gateway.


notthetup:
@AnneThKu raised a good point, to add the test case for testing `AgentID.get(null)`. I have added that.

PR #138, opened 2020-05-31
Extending CI to multiple OS

notthetup:


notthetup:
Interesting.. On this `macos-lastest` that error with `testParam1` test is much easily reproducible. Might be a good idea to instrument the test on a branch and look at the logs.

notthetup:
Finally, a slightly detailed issue caught by CI. I'm guessing it's timeout-related. 

Any idea why the tests don't fail at the first "warning", but keep a track of the number of warnings and `assert(warnings == 0)`? This makes heisenbugs harder to track since the stack trace only shows the assert. @mchitre

mchitre:
They don't fail at `assert (warnings == 0)`, but rather at `assert (warnings < 3)`. I agree this is not ideal for Heisenbugs, but it is required because the `PoissonBehavior` can overload the system with a finite probability, and then the right behavior is to drop messages. Any better ideas of how to test this?

notthetup:
It makes sense to count the warnings and expect them to be below a threshold if we're testing with `PoissonBehavior` which may lead to indeterminism.  

But the question here is the would that threshold be dependent on the underlying machine? A slower machine (common on free tiers of CI/CD servers) cause this test to give false negatives?

Secondly, the asserts are like this : 

```
    assertTrue(server.errors == 0);
    assertTrue(client1.errors == 0);
    assertTrue(client2.errors == 0);
    assertTrue(client3.errors == 0);
    assertTrue(client1.warnings < 3);
    assertTrue(client2.warnings < 3);
    assertTrue(client3.warnings < 3);
```

Can the `error` counts can be asserted at the location of the error?

ngyewch:
I'm coming in here cold, but I have some thoughts about the matter which might or might not help, feel free to shoot me down.

### On testing with `PoissonBehavior`

> It makes sense to count the warnings and expect them to be below a threshold if we're testing with PoissonBehavior which may lead to indeterminism.

I guess `PoissonBehavior` together with `RealTimePlatform` makes this test awkward in terms of real time required. The obvious fix would be to use `DiscreteEventSimulator` and run the simulation as long as you need. 

### Cause of failure

But I don't think this (`PoissonBehavior`) really addresses the test failure. From the logs, it's caused by too many warnings (from too many null responses). The null response is probably caused by slow response, since null is returned on timeout. From what I've seen, this is not exclusive to `ParameterReq`/`ParameterRsp`.

### On slower machines

> But the question here is the would that threshold be dependent on the underlying machine? A slower machine (common on free tiers of CI/CD servers) cause this test to give false negatives?

This does happen on slower or heavily loaded machines. What machines this gets tested on is outside our control, so this doesn't make for a good unit test. 

My suggestion is to test the 'slow' scenario separately (because at the root of it, it affects **all** tests that require request/response), otherwise we'd have no test at all. Whether we can test the 'slow' scenario in any useful fashion is a different matter. It'll probably in the form of a performance report rather than a test.

So in this case, switching to `DiscreteEventSimulator` might work, since it takes away the uncertainty that comes with slow / heavily loaded machines.

mchitre:
Agree. The tests we are using were not originally written with CI in mind, but to also find performance related issues. Some of them may work poorly on low-end machines, and a good way around it is the `DiscreteEventSimulator` as @ngyewch suggests. However, the `RealTimePlatform` also needs to be regression tested, so we cannot avoid its use in tests. We can perhaps loosen the timeouts and reduce the message production rates to be in line with the kinds of machines used for testing.

notthetup:
OK. Let's raise a separate issue for that. I'll finalize this PR, since all it does is enable running the CI on Windows and macOS as well.

ngyewch:
Strongly agree that `RealTimePlatform` requires extensive testing.

I did some further investigation on this by running those same tests with `DiscreteEventSimulator`. Discovered that the test failures happen on `DiscreteEventSimulator` while they pass on `RealTimePlatform` on my development machine. Increasing the request timeout helps improve the chances of passing.

See https://github.com/ngyewch/fjage-ci-test

The idea is to run the same test on both `RealTimePlatform` and `DiscreteEventSimulator` and having the option to disable the `RealTimePlatform` version when running in a CI environment.

See also `build.gradle.kts` for less noisy Gradle output.

mchitre:
Cool. Let me try it on my machine and then we can debug this.

PR #139, opened 2020-05-31
feat(base): fluent request API

ngyewch:


ngyewch:
Yeah, that was up for discussion. Wasn't quite sure about the name or location initially, was trying to pick something that will mesh with the existing API. 

But yes, it makes sense to have it at the top-level. I'll move it to the top level and merge.

PR #140, opened 2020-06-02
Javascript security fix

mchitre:
Fix to address [reported vulnerability](https://github.com/org-arl/fjage/network/alert/package-lock.json/diff/open). @notthetup you're in a much better position to know if this fix is OK, so please review before we merge.


notthetup:
Not real security issue for us since we use both the packages for development/testing and not deploy them in a public-facing application.

PR #147, opened 2020-10-08
fix(gateway): fix indexing in message queue get operation in fjage.c …

notthetup:
…gateway causing errors

This bug was found when long-running instances of `fjage.c` based gateways would crash with a segfault inside a `strcmp` called by `mqueue_get_any`. 

Investigating deeper it was discovered that this happened with the loop searching for messages in `mqueue` accessed memory outside the `mqueue` array because of this indexing error.

### DIFF
--- a/gateways/c/fjage.c
+++ b/gateways/c/fjage.c
@@ -256,8 +256,7 @@ static fjage_msg_t mqueue_get(fjage_gw_t gw, const char* clazz, const char* id) 
   if (gw == NULL) return NULL;
   _fjage_gw_t* fgw = gw;
   fjage_msg_t msg = NULL;
-  for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL;) {
-    i = (i + 1) % QUEUE_LEN;
+  for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL; i = (i + 1) % QUEUE_LEN) {
     if (fgw->mqueue[i] != NULL) {
       if (clazz != NULL) {
         const char* clazz1 = fjage_msg_get_clazz(fgw->mqueue[i]);
@@ -287,8 +286,7 @@ static fjage_msg_t mqueue_get_any(fjage_gw_t gw, const char** clazzes, int clazz
   _fjage_gw_t* fgw = gw;
   fjage_msg_t msg = NULL;
   if (clazzes == NULL || clazzlen < 1) return msg;
-  for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL;) {
-    i = (i+1)%QUEUE_LEN;
+  for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL; i = (i + 1) % QUEUE_LEN) {
     if (fgw->mqueue[i] != NULL) {
       const char* clazz1 = fjage_msg_get_clazz(fgw->mqueue[i]);
       if (clazz1 == NULL || ! mqueue_compare_any(clazz1, clazzes, clazzlen)) continue;
--- a/gateways/c/fjage.c
+++ b/gateways/c/fjage.c
@@ -256,8 +256,8 @@ static fjage_msg_t mqueue_get(fjage_gw_t gw, const char* clazz, const char* id) 
   if (gw == NULL) return NULL;
   _fjage_gw_t* fgw = gw;
   fjage_msg_t msg = NULL;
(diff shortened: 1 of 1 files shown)


gateways/c/fjage.c
```diff
@@ -257,6 +257,7 @@ static fjage_msg_t mqueue_get(fjage_gw_t gw, const char* clazz, const char* id)
   _fjage_gw_t* fgw = gw;
   fjage_msg_t msg = NULL;
   for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL; i++) {
+    if (i >= QUEUE_LEN) i = i%QUEUE_LEN;
```
mchitre, line 260: Would it be better to just use a `i = (i+1)%QUEUE_LEN` as part of the for loop? It avoids a condition check and is more localized in terms of `i` update.

gateways/c/fjage.c
```diff
@@ -287,13 +288,13 @@ static fjage_msg_t mqueue_get_any(fjage_gw_t gw, const char** clazzes, int clazz
   fjage_msg_t msg = NULL;
   if (clazzes == NULL || clazzlen < 1) return msg;
   for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL; i++) {
+    if (i >= QUEUE_LEN) i = i%QUEUE_LEN;
```
mchitre, line 291: Same comment as above.

gateways/c/fjage.c
```diff
@@ -287,13 +288,13 @@ static fjage_msg_t mqueue_get_any(fjage_gw_t gw, const char** clazzes, int clazz
...
     if (fgw->mqueue[i] != NULL) {
       const char* clazz1 = fjage_msg_get_clazz(fgw->mqueue[i]);
       if (clazz1 == NULL || ! mqueue_compare_any(clazz1, clazzes, clazzlen)) continue;
       msg = fgw->mqueue[i];
       fgw->mqueue[i] = NULL;
       if (i == fgw->mqueue_tail) fgw->mqueue_tail = (fgw->mqueue_tail+1) % QUEUE_LEN;
-      // break out
```
mchitre, line 296: Good to remove this, but any idea of the history of this comment, as it doesn't make sense here

notthetup: Looks like it was a comment added while development to point out the fact that if the code reached there it would automatically break out of the for-loop. I guess It was missed from being removed before PR.

https://github.com/org-arl/fjage/blame/b2b12fdb10df964c53414bdb798a29ccb9ce80b4/gateways/c/fjage.c#L296

gateways/c/fjage.c
notthetup: Updated.

gateways/c/fjage.c
notthetup: Updated.

gateways/c/fjage.c
```diff
@@ -256,7 +256,8 @@ static fjage_msg_t mqueue_get(fjage_gw_t gw, const char* clazz, const char* id)
   if (gw == NULL) return NULL;
   _fjage_gw_t* fgw = gw;
   fjage_msg_t msg = NULL;
-  for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL; i++) {
+  for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL;) {
+    i = (i + 1) % QUEUE_LEN;
```
mchitre, line 260: This should really go inside the `for`. It's not equivalent to put it here, as here it'll increment at the start of the loop rather than at the end.

gateways/c/fjage.c
```diff
@@ -286,14 +287,14 @@ static fjage_msg_t mqueue_get_any(fjage_gw_t gw, const char** clazzes, int clazz
   _fjage_gw_t* fgw = gw;
   fjage_msg_t msg = NULL;
   if (clazzes == NULL || clazzlen < 1) return msg;
-  for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL; i++) {
+  for (int i = fgw->mqueue_tail; i != fgw->mqueue_head && msg == NULL;) {
+    i = (i+1)%QUEUE_LEN;
```
mchitre, line 291: Same as above

mchitre (changes requested):

gateways/c/fjage.c
notthetup: Yikes. I get you now.

mchitre (approved):

PR #163, opened 2021-01-09
Add support for authentication & firewall

### LINKED ISSUE #162: Add support for auth & firewall in master container
- [x] Support `auth` field in `alive` JSON message
- [x] Provide plug-in interface to check auth info, with a default implementation based on flat files
- [x] Map credentials to a set of firewall rules, and apply them to incoming/outgoing messages
- [x] Provide plug-in interface to defining firewall rules, with a default implementation based on flat files

mchitre:
Closes #162 

Slight changes in protocol from original plan, for consistency:
```
{"alive": true}
{"action": "auth", "creds": "userid/password/certs"}
```

To enable firewall on all connections to a master container:
```
Firewall firewall = new AllowAfterAuth()   // or whatever firewall you want to use
MasterContainer master = new MasterContainer(platform, firewall)
```

To enable firewall only on specific connector (e.g. web sockets):
```
MasterContainer master = new MasterContainer(platform)
WebServer.getInstance(8080).add("/", "/org/arl/fjage/web")
Connector conn = new WebSocketConnector(8080, "/shell/ws")
Firewall firewall = new AllowAfterAuth()   // or whatever firewall you want to use
master.addConnector(conn, firewall)
```

To write your own authentication, extend `AbstractFirewall` class and implement `authenticate()` method. To have more control over which messages pass through and not, you can also override the `permit()` methods.

Three trivial firewalls included:
- `AllowAll` lets any messages pass through, even without authentication
- `DenyAll` always denies messages and authentication requests
- `AllowAfterAuth` requires an authentication message, but accepts any credentials


src/main/java/org/arl/fjage/remote/ConnectionHandler.java
```diff
@@ -70,6 +83,7 @@ public void run() {
         println(ALIVE);
       }
     }
+    fw.authenticate(conn, null);
```
notthetup, line 86: What does this check on the connection do here? Also `fw.authenticate` returns a true/false. So is the idea that a firewall that doesn't allow this connection will close it inside it's `authenticate` implementation?

notthetup (approved):

PR #165, opened 2021-01-22
Updating gateway doc for authentication support

notthetup:
With the added support for "auth" (Authentication/Authorization) in fjage's protocol, I have updated the Gateway.md document which documents the requirements for Gateways.

Summary : 

- added an `authenticate` method on the Gateway
- `receive` method will return either the `Message|Error|null` where the Error(AuthorizationError) can be returned if there was an `{auth: false}` when a certain message was received.
- agentForService/agentsForService may also receive an `{auth: false}` and must return `Error` in that case

Discussion : 

- How to identify which messages received an `{auth: false}`, especially if multiple messages are sent.
- When using `send`, and the messages gets an `{auth: false}` response, it will be ignored. Is that OK?

### DIFF
--- a/gateways/Gateways.md
+++ b/gateways/Gateways.md
@@ -15,7 +15,7 @@ All fjåge Gateway implementations should implement the following classes and me
 
 ### `authenticate` :: String credentials -> Boolean
 - Tries to authenticate the Gateway connection with the master container.
-- Must send the credential string using command `{"action":"auth","creds":"boo","id":""}`
+- Must send the credential string using command `{"action":"auth","creds":"boo"}`
 - Must wait for a the first `{"auth": true}` or `{"auth": false}` received from the master and return the result.
 - Must default timeout to 1000 millisecond.
 
--- a/gateways/Gateways.md
+++ b/gateways/Gateways.md
@@ -13,7 +13,13 @@ All fjåge Gateway implementations should implement the following classes and me
 - Must return a `null` if the connection to master container fails on the first attempt.
 - May support auto-reconnect, where, once a connection with the master container is established if the connection fails, the Gateway tries to reconnect automatically.
 
-### `getAgentID()` :: Void -> AgentID
+### `authenticate` :: String credentials -> Boolean
+- Tries to authenticate the Gateway connection with the master container.
+- Must send the credential string using command `{"auth": "<credentials>"}`
+- Must wait for a the first `{"auth": true}` or `{"auth": false}` received from the master and return the result.
+- Must default timeout to 1000 millisecond.
+
+### `getAgentID()` :: Void -> AgentID | Error
 -  Returns the _AgentID_ associated with the gateway.
 -  May be implemented as a property `agentID` on the _Gateway_.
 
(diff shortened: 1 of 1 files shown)


gateways/Gateways.md
```diff
@@ -13,18 +13,24 @@ All fjåge Gateway implementations should implement the following classes and me
 - Must return a `null` if the connection to master container fails on the first attempt.
 - May support auto-reconnect, where, once a connection with the master container is established if the connection fails, the Gateway tries to reconnect automatically.
 
-### `getAgentID()` :: Void -> AgentID
+### `authenticate` :: String credentials -> Boolean
+- Tries to authenticate the Gateway connection with the master container.
+- Must send the credential string using command `{"action":"auth","creds":"boo"}`
+- Must wait for a the first `{"auth": true}` or `{"auth": false}` received from the master and return the result.
+- Must default timeout to 1000 millisecond.
```
mchitre, line 20: Is 1000 ms a MUST?

gateways/Gateways.md
```diff
@@ -13,18 +13,24 @@ All fjåge Gateway implementations should implement the following classes and me
 - Must return a `null` if the connection to master container fails on the first attempt.
 - May support auto-reconnect, where, once a connection with the master container is established if the connection fails, the Gateway tries to reconnect automatically.
 
-### `getAgentID()` :: Void -> AgentID
+### `authenticate` :: String credentials -> Boolean
+- Tries to authenticate the Gateway connection with the master container.
+- Must send the credential string using command `{"action":"auth","creds":"boo"}`
```
mchitre, line 18: Do we need a comment on what `creds` may contain?

gateways/Gateways.md
```diff
@@ -13,18 +13,24 @@ All fjåge Gateway implementations should implement the following classes and me
...
 ### `close()` :: Void -> Void
 - Closes the _Gateway_.
 - Must send a `{"alive": false}` message to the master container before closing.
 
-### `send()` :: Message -> Boolean
+### `send()` :: Message -> Boolean  
 - Sends a _Message_ to the recipient indicated in the message.
 
-### `receive()` :: (Object filter), (Int timeout) -> Message
+### `receive()` :: (Object filter), (Int timeout) -> Message | Error
```
mchitre, line 33: returning `Error` makes sense for some languages, other places it is throwing error. Do we need to define it as a return value?

gateways/Gateways.md
```diff
@@ -37,6 +43,7 @@ All fjåge Gateway implementations should implement the following classes and me
 
 ### `request()` :: Message, (Int timeout) -> Message
 - Sends a request and waits for a response.
+- May return an `Error` of type `AuthorizationError` if a `{"auth": false}` is seen while waiting for the response
```
mchitre, line 46: Do we want to go update all gateways to throw exceptions?

gateways/Gateways.md
notthetup: Good point. Will change to a suggestion.

gateways/Gateways.md
notthetup: That would depend on the implementation of the Firewall that is being used.

mchitre: Yes, but just a comment on this to state so?

gateways/Gateways.md
mchitre: As discussed in #167, can't do this generally.

gateways/Gateways.md
mchitre: As discussed in #167, can't do this generally.

PR #179, opened 2021-02-09
Fjagejs events

notthetup:
Restructuring internal events [`conn`, `rx`, `rxmsg`, `txmsg`, `tx`] to be accessible to user of the fjage.js Gateways using the standard `addEventListener` `removeEventListener` interface. This makes it easy to add more events in the future without having to add new external-facing APIs.

This mechanism replaces the "observers", but keeps the external facings `addMessageListener ` , `removeMessageListener ` and `addConnListener` and `removeConnListener` APIs for now.

Also adding the playground UI in the test repository for testing fjage messages.


mchitre (approved):
If I read this correctly, the API does not change, so backward compatibility is maintained. Just the internals change by replacing the observers by a more flexible event listeners framework.

PR #189, opened 2021-04-06
Update fjagec Makefile

notthetup:
Using the now recommended `-_DEFAULT_SOURCE` flag instead of the `_BSD_SOURCE` flag


mchitre (approved):

notthetup:
Yikes. This change accidently reverted the fix that actually happened in https://github.com/org-arl/fjage/commit/cc744708220fb742e7231e68782465cd6e98eaab#diff-a0aff3bdfebe8aaf6fb2c8f0ec865796d4dcbc55a94c1ba29ce18a9753c6516a

I will raise a PR to revert this fix.

PR #212, opened 2021-06-29
Fix sporadic hanging of WebSocket Connector 

notthetup:
It was noticed that under some load when a WebSocket Client disconnects abruptly (for eg. Phone switching from 3G to WiFi) Jetty's `session.getRemote().sendString(s)` method can sometimes end up in a blocked state. Since the original design was single-threaded, a blocked `session.getRemote().sendString` means the entire WebSocketConnector was blocked.

This change moves to use `session.getRemote().sendStringByFuture`, the async version of the same API. But since here the underlying thread which gets blocked is provided by Jetty, WebSocketConnector can do a cleanup of async tasks that don't complete within the timeout and terminate those connections.

### DIFF
--- a/src/main/java/org/arl/fjage/connectors/WebSocketConnector.java
+++ b/src/main/java/org/arl/fjage/connectors/WebSocketConnector.java
@@ -262,8 +262,7 @@ public class WebSocketConnector implements Connector, WebSocketCreator {
           } catch (Exception e){
             log.warning(e.toString());
           }
-          long t = System.currentTimeMillis()-start;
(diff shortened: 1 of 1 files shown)


mchitre (approved):
Any performance impact in switching to the async API?

notthetup:
I haven't noticed any performance impact, although I have not tested it exhaustively. Do any of the standard tests cover performance? We can adapt one for WebSocket connector and verify if it works fine with this change.

Ideally, threading out the actual writes to various WebSockets could allow us to run the `send`s parallelly (that's not what we're doing here), so we might be able to get better performance using the async approach if we find out that this is affecting performance negatively.

mchitre:
Let's try it and we'll deal with it if we see a performance impact

Issue #222, opened 2021-09-06
Connection-based firewall


ngyewch:
`org.arl.fjage.auth.AbstractFirewall` does not seem to take multiple connections (with different credentials) into account. The `auth ` flag in the code essentially behaves as a global variable. Playing around with this seems to show this behaviour. I might be using this wrong though.

I've made something that addresses this issue: https://github.com/ngyewch/fjage-extras/tree/main/fjage-security

mchitre:
@ngyewch you're right that although each `ConnectionHandler` supports a different firewall instance, the current very simple implementation indeed gives the same `fw` object to each `ConnectionHandler`.

Your solution of using a thread-local session will solve the problem and will work. However, it assumes a single thread per connector, and this could turn out to be brittle in longer term if we ever have multi-threaded connectors. I can think of two solutions:
1. Pass the connector as a parameter to `permit()`, allowing it to make decisions based no connector instance.
2. Have a `fw` per connection and use a factory method to create a new `fw` each time a connection is opened.

Preferences or thoughts on this?

ngyewch:
> Your solution of using a thread-local session will solve the problem and will work. However, it assumes a single thread per connector, and this could turn out to be brittle in longer term if we ever have multi-threaded connectors.

Agreed. It's a hack for something I needed to implement for use by a 3rd-party (they're not directly using the firewall, just connecting to the `MasterContainer`). 

> 1. Pass the connector as a parameter to permit(), allowing it to make decisions based no connector instance.

From a pure design standpoint, `Firewall` should not need access to `Connector`.

> 2. Have a fw per connection and use a factory method to create a new fw each time a connection is opened.

This probably makes more sense.

```
public interface FirewallFactory {
    Firewall newFirewall();
}
```

```
MasterContainer(Platform platform, int port, FirewallFactory fwFactory)
```

In this case `Firewall.authenticate()` shouldn't require a `Connector` parameter (ideally), but this can't happen because of backward-compatibility.

mchitre:
We can rework the firewall API, as the API isn't used much anywhere yet and can be considered experimental. Better to get the design right now, than to hack around later.

PR #223, opened 2021-09-09
Auth/firewall revamp

ngyewch:
This revamp of auth/firewall makes it easier to differentiate between connections.

See #222

As discussed, this PR includes breaking/signature changes in the following classes:
* `org.arl.fjage.auth.AllowAfterAuth`
* `org.arl.fjage.auth.AllowAll`
* `org.arl.fjage.auth.DenyAll`
* `org.arl.fjage.auth.Firewall`
* `org.arl.fjage.remote.MasterContainer`
    * some constructors

I've also had to relax the scope of:
* `org.arl.fjage.remote.Action`

### DIFF
--- a/src/main/java/org/arl/fjage/remote/ConnectionHandler.java
+++ b/src/main/java/org/arl/fjage/remote/ConnectionHandler.java
@@ -95,7 +95,6 @@ class ConnectionHandler extends Thread {
         println(ALIVE);
       }
     }
-    fw.authenticate(conn, null);
     while (conn != null) {
       String s = null;
       try {
@@ -144,7 +143,7 @@ class ConnectionHandler extends Thread {
         } else {
           // new request
           if (rq.action == Action.AUTH) {
-            boolean b = fw.authenticate(conn, rq.creds);
+            boolean b = fw.authenticate(rq.creds);
             respondAuth(rq, b);
           }
           else if (fw.permit(rq)) pool.execute(new RemoteTask(rq));
@@ -154,7 +153,6 @@ class ConnectionHandler extends Thread {
         log.warning("Bad JSON request: "+ex.toString() + " in " + s);
       }
     }
-    fw.authenticate(conn, null);
     close();
     pool.shutdown();
   }
--- a/src/main/java/org/arl/fjage/remote/ConnectionHandler.java
+++ b/src/main/java/org/arl/fjage/remote/ConnectionHandler.java
@@ -153,6 +153,7 @@ class ConnectionHandler extends Thread {
         log.warning("Bad JSON request: "+ex.toString() + " in " + s);
       }
     }
+    fw.signoff();
     close();
     pool.shutdown();
   }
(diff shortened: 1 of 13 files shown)


src/main/java/org/arl/fjage/remote/ConnectionHandler.java
```diff
@@ -154,7 +153,6 @@ public void run() {
         log.warning("Bad JSON request: "+ex.toString() + " in " + s);
       }
     }
-    fw.authenticate(conn, null);
```
mchitre, line 157: Do we need a "logout" call which allows a Firewall to cleanup things when the connection is no longer used?

mchitre (approved):
Now I recall how the original design was supposed to work. The binding to the connector was supposed to allow the stateful behavior per connection. Instead of a thread-local variable, the idea was to use a map with the connection to store state. However, I do prefer this stateful design more, so I propose we go with the change.

src/main/java/org/arl/fjage/remote/ConnectionHandler.java
ngyewch: Yeah, I was considering that and sort of forgot about it. Yeah, it's probably good to have some way to clean this up.

I guess we could put that back in, only problem I have with it is that it is currently indistinguishable from `SlaveContainer.authenticate(null)`. Maybe I'm overthinking it? 

Perhaps something like `Firewall.discard()`. Overkill?

Either way is fine, just presenting some options.

mchitre: `Firewall.discard()` (or my preference for `Firewall.signoff()` or something similar) works well.

src/test/java/org/arl/fjage/test/auth/SimpleFirewallSupplier.java
```diff
@@ -0,0 +1,129 @@
...
+import org.arl.fjage.auth.Firewall;
+import org.arl.fjage.remote.Action;
+import org.arl.fjage.remote.JsonMessage;
+
+import java.util.*;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+public class SimpleFirewallSupplier
+    implements Supplier<Firewall> {
```
ngyewch, line 23: @mchitre `Firewall` as it stands is pretty low-level. We might we to consider including something like `SimpleFirewallSupplier` that addresses common usage. If you agree, we will need to review this class.

src/main/java/org/arl/fjage/remote/ConnectionHandler.java
ngyewch: Implemented in 02fd3a7d133b306b9f28a3f06176946d5ef26336

src/test/java/org/arl/fjage/test/auth/SimpleFirewallSupplier.java
mchitre: Yes, sure, having a basic implementation that is useful is nice. We might want to clearly articulate what the implementation is meant to do and not, and if someone had to extend it to link with authentication systems, what it would take.

mchitre: Might it be better to have just this implementation as a separate PR? Merge the framework first, and then add in a simple implementation?

PR #224, opened 2021-09-10
SimpleFirewallSupplier

ngyewch:
Following up from https://github.com/org-arl/fjage/pull/223#pullrequestreview-750426968

`SimpleFirewallSupplier` has the following main objectives

* Restrict access on a per-service, -agent, -topic basis
* Provide basic per-user configuration
* Provide basic provisioning methods


mchitre (approved):

src/main/java/org/arl/fjage/auth/SimpleFirewallSupplier.java
```diff
@@ -0,0 +1,335 @@
+/******************************************************************************
+
+Copyright (c) 2013, Mandar Chitre
```
mchitre, line 3: You probably want to change this line :-)

notthetup (approved):

PR #240, opened 2022-03-16
Added required header file in fjage.c for signal

prasadtiru:


notthetup:
Would this cause the windows build to fail?

notthetup:
OK. I tested. It doesn't cause any issues.

notthetup (approved):

Issue #253, opened 2022-08-30
Listing of an `AgentID` with a subclass of `NamedParameter` causes a `MissingPropertyException`


notthetup:
When using subclasses of `org.arl.fjage.param.NamedParameter` which are exposed by an Agent, when listing the `AgentID` in a Groovy shell (like the one provided by the ShellAgent), a `MissingPropertyException` is thrown.

For example,`org.arl.fjage.sentuator.ConfigParam` from [fjage-sentuator](https://github.com/org-arl/fjage-sentuator) is a subclass of `org.arl.fjage.param.NamedParameter`. 

https://github.com/org-arl/fjage-sentuator/blob/master/src/main/groovy/org/arl/fjage/sentuator/ConfigParam.groovy

```
mysentuator
```

When but when listing an `Agent` which exposes a `ConfigParam`, like say a `Sentuator` on a groovy shell, nothing gets printed on the shell and an Exception is logged.

```
1661770440607|WARNING|org.arl.fjage.shell.ShellAgent/audio@48:lambda$run$0|Exec failure: groovy.lang.MissingPropertyException: No such property: name for class: org.arl.fjage.sentuator.ConfigParam
```

This is due to `AgentID.metaClass.getAllParameters` in [GroovyExtensions](https://github.com/org-arl/fjage/blob/master/src/main/groovy/org/arl/fjage/GroovyExtensions.groovy#L73) using the `.name` field on the `NamedParameter`. This works for all `NamedParameter` since the field exists (but is private) so [Groovy allows access to it](https://issues.apache.org/jira/browse/GROOVY-3010) but since `ConfigParam` is defined in another Jar, it fails.

The fix should be to use the `.name()` method instead.

PR #254, opened 2022-08-31
Use the name() method instead of the name field in AgentID.metaClass.getAllParameters

notthetup:
Using the `.name()` method instead of the `.name` field to get the name of a `NamedParameter` in `AgentID.metaClass.getAllParameters`

Fix https://github.com/org-arl/fjage/issues/253


notthetup:
OK. This doesn't completely solve it.

It now shows this : 

```
 gps
« GPSD Sentuator »

[NamedParameter]
  tickInterval ⤇ 0
  updateInterval = 5000

[org.arl.fjage.sentuator.SentuatorParam]
  enable = true
  poll = 5000
```

mchitre:
Showing `[NamedParameter]` is the correct behavior. The parameter does not belong to the `org.arl.fjage.sentuator.ConfigParam` class and should not be qualified using this.

mchitre (approved):

PR #263, opened 2022-12-29
Add syntax highlighting for fjåge log files

mchitre:
To install, simply `Cmd-shift-P` and choose "Developer: Install extension from location..." and select the `vscode` folder. Then open the log file and select `fjage-log` in the language selection in the status bar at the bottom-right.

Example screenshot:
image


notthetup (approved):

notthetup:
We should document this somewhere externally. Maybe a readme in the vscode directory?

mchitre:
Yes, we should provide instructions for use, as shown above. A short README would be good, once we determine that this was good enough for public consumption. For now, I'd consider it beta testing, as I've just about started using it.

PR #264, opened 2023-01-09
Fix get_param_str in fjage.c

notthetup:
This is a bug fix for a bug in `fjage.c` where `fjage_get_param_str` causes a null pointer to be returned even if there is a valid string available.

The reason is : 

https://github.com/org-arl/fjage/blob/master/gateways/c/src/fjage.c#L1365..L1366
`v = fjage_msg_get_string(msg, "value");`
`v` is a pointer into the message's body to the string..
And then we free the message body with `fjage_msg_destroy(msg);`
So `v` is now pointing into freed memory.

The fix is to return a copy of the string instead.

### DIFF
--- a/gateways/c/src/fjage.c
+++ b/gateways/c/src/fjage.c
@@ -1365,7 +1365,10 @@ int fjage_param_get_string(fjage_gw_t gw, fjage_aid_t aid, const char* param, in
     if (fjage_msg_get_performative(msg) == FJAGE_INFORM) {
       const char* v = fjage_msg_get_string(msg, "value");
       if (v != NULL){
-        if (strval != NULL && len > 0) rv = strlcpy((char*)strval, v, len);
+        if (strval != NULL && len > 0) {
+          strncpy((char*)strval, v, len);
+          rv = strlen(v) > len ? len : strlen(v);
+        }
         else rv = strlen(v);
       }
     }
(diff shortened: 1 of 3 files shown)


mchitre:
This is a BREAKING change, since it changes the API and leads to memory leaks for all previous users of fjåge! We need to tag the release as such and mention it clearly in the release notes. Are there any other APIs we need to break? If so, we should do it at the same time.

notthetup:
True. I am not sure if there are any other breaking changes pending for the fjage.c Gateway.

P.S. Also happy to change this fix in a way that doesn't break the API, but I couldn't think of any other approach.

notthetup:
I do prefer that style of API. It makes memory ownership more transparent. But that'll also be a breaking change.

Also, if that's the approach we're going for, we should use that style across the board for other APIs which return pointers etc.

mchitre:
It's better breaking, in the sense that it'll fail at compilation and force a user to make changes with understanding. The other breaking is dangerous as the compilation will succeed and they'll get memory leaks silently.

mchitre (changes requested):
Requested changes as per discussion on the PR

mchitre (approved):
Might want to consider adding test cases for checking that it returns required buffer size correctly if NULL is passed in.

notthetup:
If `NULL` is passed in for `const char* strval`, then ideally it shouldn't copy ANY characters into the buffer. In such a scenario, should it return `0`?

mchitre:
It should return the actual length so user knows what size buffer needs to be passed in.

mchitre (approved):

PR #265, opened 2023-01-09
Updating dependencies

notthetup:
As a first step to upgrading to newer Groovy/Java, this upgrades all the dependencies to the latest of MINOR/PATCH.

I verified that jSerialComm which had broken when we upgraded to jSerialComm 2.9.2 (https://github.com/org-arl/fjage/commit/624a3c291c2447d5d04ea6db7fbf0d085e0b4365) has been fixed in jSerialComm v2.9.3 (https://github.com/Fazecast/jSerialComm/releases/tag/v2.9.3) and doesn't break modem builds.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,18 +38,18 @@ repositories {
 }
 
 dependencies {
-  api 'org.codehaus.groovy:groovy:2.5.9'
-  api 'org.jline:jline:3.12.1'
-  api 'org.apache.commons:commons-lang3:3.9'
-  api 'commons-io:commons-io:2.6'
+  api 'org.codehaus.groovy:groovy:2.5.20'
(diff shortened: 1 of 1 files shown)


notthetup:
Also, I do remember @ngyewch mentioned something about higher versions of Groovy 2.5 breaking something either in fjåge or in one of the downstream projects. But I couldn't reproduce it. Do you remember what it was that broke @ngyewch ?

ngyewch:
> Also, I do remember @ngyewch mentioned something about higher versions of Groovy 2.5 breaking something either in fjåge or in one of the downstream projects. But I couldn't reproduce it. Do you remember what it was that broke @ngyewch ?

I cannot recall, and I cannot seem to replicate it either.

mchitre (approved):

PR #273, opened 2023-02-19
Migrate to Github flow model

notthetup:
Proposing to move the Github flow model where we'll work on the `master` and release from there. https://docs.github.com/en/get-started/quickstart/github-flow


mchitre (approved):
May need to change base branch for open PRs

manuignatius:
Possible to have a discussion about the pros and cons of this model? I looked at the github docs, its the same workflow we follow now when working off dev rt?

PR #276, opened 2023-03-21
Checking if the executor pool is shutdown before executing

notthetup:
ConnectionHandler uses `SingleThreadExecutor` to send messages on a given Connector. If the Connector stops, it can cause the ExecutorPool to be stopped. So we need to check if the Executor has not stopped before trying to send something on the connector.

### DIFF
--- a/src/main/java/org/arl/fjage/remote/ConnectionHandler.java
+++ b/src/main/java/org/arl/fjage/remote/ConnectionHandler.java
@@ -160,6 +160,13 @@ class ConnectionHandler extends Thread {
     fw.signoff();
     close();
     pool.shutdown();
+    try {
+      if (!pool.awaitTermination(200, TimeUnit.MILLISECONDS)) {
+        pool.shutdownNow();
+      }
+    } catch (InterruptedException e) {
+      pool.shutdownNow();
+    }
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


notthetup:
This was happening when a `SlaveContainer` stopped responding and the connection to the `MasterContainer` closed. If some Agent was still trying to send a response to some request from the `SlaveContainer`, it could cause an exception.

```
1679368369915|INFO|org.arl.fjage.remote.MasterContainer@47:connectionClosed|Connection tcp:///127.0.1.1:1100//127.0.0.1.37286 closed
1679368369941|SEVERE|org.arl.unet.nodeinfo.NodeInfo/audio@44:run|Exception in agent: node

java.util.concurrent.RejectedExecutionException: Task org.arl.fjage.remote.ConnectionHandler$$Lambda$236/1911266333@50ecfb8c rejected from java.util.concurrent.ThreadPoolExecutor@50a52d2b[Shutting down, pool size = 1, active threads = 1, queued tasks = 1, completed tasks = 2]
Stack trace: ...
   org.arl.fjage.remote.ConnectionHandler.printlnQueued(ConnectionHandler.java:229)
   org.arl.fjage.remote.MasterContainer.send(MasterContainer.java:235)
   org.arl.fjage.remote.MasterContainer.send(MasterContainer.java:216)
   org.arl.fjage.Agent.send(Agent.java:411)
    ...
```

PR #278, opened 2023-04-12
Expose a `shellready` variable on the webshell iframe

notthetup:
`shellready` will be set to `false` at initialization and only be set to `true` when the first message is received from the WebSocket connection.


mchitre:
What happens if connection is lost? Should shell go not ready?

PR #285, opened 2023-08-07
Allow named params to match with param enums in `get()`

### LINKED ISSUE #284: AgentID.get(NamedParameter) does not work
This works:
```groovy
> req = new ParameterReq(node).get(new NamedParameter("nodeName"))
ParameterReq[nodeName:?]

> node.request(req)
ParameterRsp[nodeName:A]
```

This doesn't:
```groovy
> node.get(new NamedParameter("nodeName"))
```

Based on my understanding of the code, the two should be exactly the same:
https://github.com/org-arl/fjage/blob/454ca543e83c82cf7e8cfc0632c28b7a1cd55a92/src/main/java/org/arl/fjage/AgentID.java#L189-L214

mchitre:
Fixes #284

After fix:
```groovy
> node.nodeName
Barrackpore.local
> import org.arl.fjage.param.*
> req = new ParameterReq(node).get(new NamedParameter("nodeName"))
ParameterReq[nodeName:?]
> x = node.request(req)
ParameterRsp[nodeName:Barrackpore.local]
> x.get(org.arl.unet.nodeinfo.NodeInfoParam.nodeName)
Barrackpore.local
> x.get(new NamedParameter("nodeName"))
Barrackpore.local
> node.get(new NamedParameter("nodeName"))
Barrackpore.local
> req = new ParameterReq(node)
ParameterReq[]
> x = node.request(req)
ParameterRsp[address:99 origin:[D@10ef0353 canForward:true heading:null depth:null description*:Manages and maintains node information and attributes. nodeName:Barrackpore.local title*:Node information pitch:null speed:null time*:Tue Aug 08 15:48:44 SGT 2023 roll:null addressSize:8 turnRate:null location:[D@2911ab14 diveRate:null mobility:false]
> x.get(org.arl.unet.nodeinfo.NodeInfoParam.nodeName)
Barrackpore.local
> x.get(new NamedParameter("nodeName"))
Barrackpore.local
```

### DIFF
--- a/src/main/java/org/arl/fjage/param/NamedParameter.java
+++ b/src/main/java/org/arl/fjage/param/NamedParameter.java
@@ -74,20 +74,22 @@ public class NamedParameter implements Parameter, Serializable {
   }
 
   /**
-   * Compares if two named parameters are equals.
+   * Compares if a parameter is equivalent to the named parameter.
    *
    * @return true if two parameters are equal
    */
   @Override
   public boolean equals(Object obj) {
-    if (!(obj instanceof NamedParameter)) return false;
-    NamedParameter p = (NamedParameter)obj;
+    if (!(obj instanceof Parameter)) return false;
     String name1 = name;
     if (name1.contains(".")) name1 = name1.substring(name1.indexOf('.')+1);
-    String name2 = p.name;
+    String name2 = obj.toString();
     if (name2.contains(".")) name2 = name2.substring(name2.indexOf('.')+1);
     if (!name1.equals(name2)) return false;
-    if (ord >= 0 && p.ord >= 0 && p.ord != ord) return false;
+    if (obj instanceof NamedParameter) {
+      NamedParameter p = (NamedParameter)obj;
+      if (ord >= 0 && p.ord >= 0 && p.ord != ord) return false;
+    }
     return true;
   }
 
--- a/src/main/java/org/arl/fjage/param/NamedParameter.java
+++ b/src/main/java/org/arl/fjage/param/NamedParameter.java
@@ -74,22 +74,20 @@ public class NamedParameter implements Parameter, Serializable {
   }
 
   /**
-   * Compares if a parameter is equivalent to the named parameter.
+   * Compares if two named parameters are equal.
    *
    * @return true if two parameters are equal
    */
   @Override
   public boolean equals(Object obj) {
-    if (!(obj instanceof Parameter)) return false;
+    if (!(obj instanceof NamedParameter)) return false;
+    NamedParameter p = (NamedParameter)obj;
     String name1 = name;
     if (name1.contains(".")) name1 = name1.substring(name1.indexOf('.')+1);
-    String name2 = obj.toString();
+    String name2 = p.name;
     if (name2.contains(".")) name2 = name2.substring(name2.indexOf('.')+1);
     if (!name1.equals(name2)) return false;
-    if (obj instanceof NamedParameter) {
-      NamedParameter p = (NamedParameter)obj;
-      if (ord >= 0 && p.ord >= 0 && p.ord != ord) return false;
-    }
+    if (ord >= 0 && p.ord >= 0 && p.ord != ord) return false;
     return true;
   }
 
--- a/src/main/java/org/arl/fjage/param/Parameter.java
+++ b/src/main/java/org/arl/fjage/param/Parameter.java
@@ -22,4 +22,11 @@ public interface Parameter {
    */
   public int ordinal();
 
+  /**
+   * Gets the parameter name.
+   *
+   * @return unqualified parameter name
+   */
+  public String name();
+
 }
--- a/src/main/java/org/arl/fjage/param/ParameterRsp.java
+++ b/src/main/java/org/arl/fjage/param/ParameterRsp.java
@@ -90,10 +90,10 @@ public class ParameterRsp extends Message {
    * @return value of requested parameter
    */
   public Object get(Parameter param) {
-    if (param == null) return null;
-    if (this.param == null) return null;
+    if (this.param == null || param == null) return null;
+    param = resolve(param);
     Object rv = null;
-    if (param.equals(this.param)) {
+    if (this.param.equals(param)) {
       if (value == null) return null;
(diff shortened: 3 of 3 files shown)


src/main/java/org/arl/fjage/param/NamedParameter.java
```diff
@@ -74,20 +74,22 @@ public String toString() {
   }
 
   /**
-   * Compares if two named parameters are equals.
+   * Compares if a parameter is equivalent to the named parameter.
    *
    * @return true if two parameters are equal
    */
   @Override
   public boolean equals(Object obj) {
```
ettersi, line 82: This implementation of `equals()` is not an equivalence relation: we have `new NamedParameter("address").equals(NodeInfoParam.address)` but `!NodeInfoParam.address.equals(new NamedParameter("address"))`, so the relation is not symmetric. This may lead to unexpected behaviour e.g. if you stick parameters into a set or map. Instead of overloading `equals()`, we should move this method into a separate function.

src/main/java/org/arl/fjage/param/NamedParameter.java
```diff
@@ -74,20 +74,22 @@ public String toString() {
...
-    if (!(obj instanceof NamedParameter)) return false;
-    NamedParameter p = (NamedParameter)obj;
+    if (!(obj instanceof Parameter)) return false;
     String name1 = name;
     if (name1.contains(".")) name1 = name1.substring(name1.indexOf('.')+1);
```
ettersi, line 85: This only peels of the first item of the path. 

```groovy
groovy:000> name = "org.arl.unet.nodeinfo.NodeInfoParams.address"

groovy:000> name.substring(name.indexOf('.')+1)
===> arl.unet.nodeinfo.NodeInfoParams.address
```

Isn't this what we want here?

```groovy
groovy:000> name.substring(name.lastIndexOf('.')+1)
===> address
```

src/main/java/org/arl/fjage/param/NamedParameter.java
```diff
@@ -74,20 +74,22 @@ public String toString() {
...
-    String name2 = p.name;
+    String name2 = obj.toString();
     if (name2.contains(".")) name2 = name2.substring(name2.indexOf('.')+1);
     if (!name1.equals(name2)) return false;
```
ettersi, line 88: I assume the intention here is that this line compares unqualified names. However, that would mean that e.g. `NodeInfoParam.address` and `CellMobileParam.address` would compare equal.

src/main/java/org/arl/fjage/param/NamedParameter.java
mchitre: Introducing a new method will entail special handling of `NamedParameter` in `ParameterRsp`, as general `Parameter` (essentially an `Enum`) will not provide an implementation. In that case, we might as well consider moving the matching logic to `ParameterRsp` and not bother with a separate method.

src/main/java/org/arl/fjage/param/NamedParameter.java
mchitre: No, I don't believe the intent here was for unqualified names. I don't recall the exact reason why this peeling off of the first segment was introduced (and I tried going back to history to reconstruct, but failed at understanding it), but I vaguely remember it had something to do with how some of the gateways dealt with parameters.

Non-named parameters are always compared fully qualified.

ettersi: I believe currently the behaviour is like this:

- [x] Enum item to enum item is compared using the full path. 
- [ ] Enum item to `NamedParameter` is always `false`.
- [ ] Fully qualified `NamedParameter` to enum item is "mostly" `false`. (Comparison is done based on `toString()`, and enums' `toString()` method returns only the item name but not the path. The "mostly" is because currently we only peel off the first component of the path.)
- [x] Unqualified `NamedParameter` to enum item is compared based on only the name. 
- [ ] `NamedParameter` to `NamedParameter` is compared based on all but the first component of the path. 

The ticks indicate behaviour as I would expect it.

src/main/java/org/arl/fjage/param/NamedParameter.java
mchitre: We don't use qualified named parameters, so this doesn't make sense. Since I don't recall the exact context in which this code was added (see comment below), I am reluctant to change this in this PR. If it's genuinely a bug, it should be fixed in its own PR.

ettersi: Ah, I guess I misunderstood you when we talked about this on the UnetStack call last week. I thought `NamedParameter`s should support fully qualified paths as well. 

In this case, my only concern with this PR is that we are abusing `equals()` in ways it wasn't intended for. Arguably a bit pedantic, but I'd say better be safe than sorry?

mchitre: My initial reaction was that the abuse of `equal()` makes the design simpler and so might be OK. But on deeper thought, you're correct in pointing out that there may be issues with Sets and Maps, as there are no guarantees of which order the comparisons will be made. I have reverted back the `equals()` to only handle named parameter comparisons and return `false` for other comparisons.

src/main/java/org/arl/fjage/param/Parameter.java
```diff
@@ -22,4 +22,11 @@ public interface Parameter {
    */
   public int ordinal();
 
+  /**
+   * Gets the parameter name.
+   *
+   * @return unqualified parameter name
+   */
+  public String name();
```
ettersi, line 30: Doesn't this require updating every single `Parameter` implementation in fjage, UnetStack, etc? Sounds like a lot of work (and pain). Alternatively, we could change the definition of `toString()` to say that `toString()` has to be the unqualified parameter name. This is trivially true for all enum parameters, and then we can put the "remove `XXX.`" logic into `NamedParameter.toString()`.

mchitre: No change needed. All parameters are enums and all enums define `name()`. Adding it to the `Parameter` interface simply exposes it without having to cast.

ettersi (approved):

PR #294, opened 2023-11-17
Fix Gateway <-> Wormhole interop

ettersi:
Consider a setup where a Java `Gateway` is connected to a master container with a Unet `Wormhole` agent running in it. In this scenario, the gateway will currently not receive any messages addressed to it. 

The technical reasons behind this are as follows. 

- The gateway's name is of the form `Gateway@<number>`, i.e. it looks just like an agent that is wormholed to the master container. 
- The Java gateway currently does not submit a `wantsMessagesFor` to the master container. Therefore, it receives all the messages that aren't claimed on the master container, but it also receives only those. 
- Consequently, the gateway doesn't receive messages addressed to it: the wormhole claims all messages addressed to `<name>@<number>`, and the master container doesn't know that the gateway would like to receive messages addressed to `Gateway@<number>`. 

This PR fixes both of these aspects. 

- It makes sure that Java `Gateway`s do send a `wantsMessagesFor` on startup. 
- It changes the gateway naming pattern from `Gateway@(\d+)` to `Gateway-(\d+)` to avoid confusion with wormholed agents. 

Strictly speaking, either change would have been enough on its own, but there's no reason not to fix both for extra robustness (and performance in the case of the `wantsMessagesFor`).

### DIFF
--- a/src/main/java/org/arl/fjage/Container.java
+++ b/src/main/java/org/arl/fjage/Container.java
@@ -569,7 +569,7 @@ public class Container {
    * Starts the container and all agents in it.
    * This should be called after init().
    */
-  void start() {
+  protected void start() {
     if (!running) {
       log.info("Starting container...");
       running = true;
--- a/src/main/java/org/arl/fjage/remote/Gateway.java
+++ b/src/main/java/org/arl/fjage/remote/Gateway.java
@@ -57,7 +57,7 @@ public class Gateway implements Messenger, Closeable {
    * @param port TCP port to connect to.
    */
   public Gateway(Platform platform, String hostname, int port) throws IOException {
-    container = new SlaveContainer(platform, "Gateway@"+hashCode(), hostname, port);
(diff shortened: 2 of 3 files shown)


mchitre (approved):

notthetup:
Do we need to update the other language Gateways? And maybe update the Gateway spec?

mchitre:
Good question. Need to check the naming of the gateways from other languages.

ettersi:
@mchitre Any chance we can make a release with this change? ST needs this for Aquarius, and making a release is probably the cleanest way to ship this code to them.

mchitre:
Yes, let me try to do that sometime today

mchitre:
Released `fjage-1.12.1`

ettersi:
> Do we need to update the other language Gateways? And maybe update the Gateway spec?

JS, Python and Julia gateways are fine. I've fixed the C gateway in https://github.com/org-arl/fjage/pull/295.

PR #295, opened 2023-11-20
Fix gateway names to avoid clashes with wormhole agent

ettersi:
- Changes the agent ID of the C gateway from `CGatewayAgent@...` to `CGateway-...` to avoid clashes with the Unet wormhole agent (see https://github.com/org-arl/fjage/pull/294). 
- Factors out the agent ID creation in the Java gateway so all ways of creating a Java gateway lead to the same agent ID.

### DIFF
--- a/gateways/c/src/fjage.c
+++ b/gateways/c/src/fjage.c
@@ -550,7 +550,7 @@ fjage_gw_t fjage_tcp_open(const char* hostname, int port) {
   fgw->intr = 0;
 #endif
   char s[64];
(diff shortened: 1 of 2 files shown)


mchitre (approved):

ettersi:
Whoops, turns out I also need the fix to `getAgentID()` introduced here for Aquarius. Can I bother you for one more release, @mchitre? Really sorry about this. I've tested that this version works, so this should be the last one 🤞

mchitre:
No problem, shortly...

mchitre:
Released fjage `1.12.2`

PR #298, opened 2024-01-16
Add option to reject various APIs and added new agents API

notthetup:
- Added a `Gateway.agents()` API to get a list of AgentIDs available on the platform
- Added a `Gateway.containsAgent()` API to check if a given AgentID is available on the platform
- Updated the following APIs to reject Promises if unable to get a response for a given Request instead of just returning `null`.
  - `Gateway.agentForService`
  - `Gateway.agentsForService`
  - `AgentID.get`
  - `AgentID.set`
- For backward compatibility, this functionality is disabled by default but it can be enabled using `Gateway({returnNullOnError: true})` in the Gateway constructor. This will be the default in some future releases.

---

With this feature, parameter set/get can be done using the construct 

```
t = await device.get("time").catch(() => {console.warn("couldn't get time")})
```

### DIFF
--- a/gateways/js/src/fjage.js
+++ b/gateways/js/src/fjage.js
@@ -728,7 +728,11 @@ export class Gateway {
   async agentForService(service) {
     let rq = { action: 'agentForService', service: service };
     let rsp = await this._msgTxRx(rq);
-    if (!rsp || !rsp.agentID) return;
+    if (!rsp) {
+      if (this._returnNullOnError) return null;
+      else throw new Error('Unable to get agent for service');
+    }
+    if (!rsp.agentID) return null;
     return new AgentID(rsp.agentID, false, this);
   }
 
@@ -742,7 +746,11 @@ export class Gateway {
     let rq = { action: 'agentsForService', service: service };
     let rsp = await this._msgTxRx(rq);
     let aids = [];
-    if (!rsp || !Array.isArray(rsp.agentIDs)) return aids;
+    if (!rsp) {
+      if (this._returnNullOnError) return aids;
+      else throw new Error('Unable to get agents for service');
+    }
+    if (!Array.isArray(rsp.agentIDs)) return aids;
     for (var i = 0; i < rsp.agentIDs.length; i++)
       aids.push(new AgentID(rsp.agentIDs[i], false, this));
     return aids;
--- a/gateways/js/src/fjage.js
+++ b/gateways/js/src/fjage.js
@@ -132,14 +132,17 @@ export class AgentID {
     msg.index = Number.isInteger(index) ? index : -1;
     const rsp = await this.owner.request(msg, timeout);
     var ret = Array.isArray(params) ? new Array(params.length).fill(null) : null;
-    if (!rsp || rsp.perf != Performative.INFORM || !rsp.param) return ret;
+    if (!rsp || rsp.perf != Performative.INFORM || !rsp.param) {
+      if (this.owner._returnNullOnError) return ret;
+      else throw new Error(`Unable to set ${this.name}.${params} to ${values}`);
+    }
     if (Array.isArray(params)) {
       if (!rsp.values) rsp.values = {};
       if (rsp.param) rsp.values[rsp.param] = rsp.value;
       const rvals = Object.keys(rsp.values);
       return params.map( p => {
         let f = rvals.find(rv => rv.endsWith(p));
-        return f ? rsp.values[f] : null;
+        return f ? rsp.values[f] : undefined;
       });
     } else {
       return rsp.value;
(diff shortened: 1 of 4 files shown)


gateways/js/src/fjage.js
```diff
@@ -132,14 +132,17 @@ export class AgentID {
...
+      if (this.owner._returnNullOnError) return ret;
+      else throw new Error(`Unable to set ${this.name}.${params} to ${values}`);
+    }
     if (Array.isArray(params)) {
       if (!rsp.values) rsp.values = {};
       if (rsp.param) rsp.values[rsp.param] = rsp.value;
       const rvals = Object.keys(rsp.values);
       return params.map( p => {
         let f = rvals.find(rv => rv.endsWith(p));
-        return f ? rsp.values[f] : null;
```
ettersi, line 142: Is there any practical difference between `null` and `undefined`? Just curious...

ettersi (commented):
> * Updated the following APIs to reject Promises if unable to get a response for a given Request instead of just returning `null`

Love this! Ideally we'd do the same also in other gateways, I suspect this would lead to much more robust code and simplify debugging. 

> * For backward compatibility, this functionality is disabled by default but it can be enabled using `Gateway({returnNullOnError: true})` in the Gateway constructor. This will be the default in some future releases.

Maybe backwards compatibility is better addressed using semantic versioning? Supporting both options in the same code is a maintenance issue, and changing the default is going to be breaking anyway. 

> * returnNullOnError

I find this name somewhat confusing, probably because my first thought upon reading this was that "Error" would refer to the "error" in "throw error instead of returning null". Maybe `returnNullOnFailedResponse`?

gateways/js/src/fjage.js
notthetup: I think the original language definition was : 

> `undefined` means a variable has been declared but has not yet been assigned a value
> `null` is an assignment value. It can be assigned to a variable as a representation of no value 

From : https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript

So you're right,  in this scenario we should return `undefined`.

notthetup:
> Love this! Ideally we'd do the same also in other gateways, I suspect this would lead to much more robust code and simplify debugging.

The major improvement here is the disambiguation between a failed `agent.parameter` request and one where the parameter value is actually `null`. Similarly for `agentForService` and other operations.

It would be great to extend this to other gateways, but it's slightly easier with JavaScript since it doesn't have blocking operations, so we have to wrap them in a `Promise`. 

In other languages we'll have to find the proper native (hopefully) mechanism to do this over blocking calls and not make it too verbose and awkward to use.

> Maybe backwards compatibility is better addressed using semantic versioning? Supporting both options in the same code is a maintenance issue, and changing the default is going to be breaking anyway.

Yes. Agreed. I thought about this. But one awkward issue here is that we bundle `fjage.js` into `fjage.jar` as well, so that would require us semver upgrade `fjage.jar` as well, since we don't want a user to upgrade fjage.jar and automatically get a new `fjage.js` which might break things with their app. I haven't found a nice way to pin dependency between fjage.js and fjage.jar.

> I find this name somewhat confusing, probably because my first thought upon reading this was that "Error" would refer to the "error" in "throw error instead of returning null". Maybe returnNullOnFailedResponse?

`returnNullOnFailedResponse` sounds good. I'll update.

ettersi:
> > Maybe backwards compatibility is better addressed using semantic versioning? Supporting both options in the same code is a maintenance issue, and changing the default is going to be breaking anyway.
> 
> Yes. Agreed. I thought about this. But one awkward issue here is that we bundle `fjage.js` into `fjage.jar` as well, so that would require us semver upgrade `fjage.jar` as well, since we don't want a user to upgrade fjage.jar and automatically get a new `fjage.js` which might break things with their app. I haven't found a nice way to pin dependency between fjage.js and fjage.jar.

If `fjage.js` is part of `fjage.jar`, then it shouldn't be surprising that a breaking update for `fjage.jar` also breaks `fjage.js`, no? I do see the issue that this would force us to update all of `unet-web` at once, though.

notthetup:
> I do see the issue that this would force us to update all of `unet-web` at once, though.

Also, we won't be able to pick up any new functionality in fjåge (like https://github.com/org-arl/fjage/pull/300 or https://github.com/org-arl/fjage/pull/301) until the entire `unet-web` updated to use the new API.

ettersi:
> Hence, I thought of using the "deprecated" approach and hopefully removing the original functionality it in a future release when we have the chance to up-rev fjage.jar as well.

Sounds reasonable 👍

ettersi (approved):

PR #305, opened 2024-02-20
Allow the WebShell dark mode toggle to be disabled

notthetup:
We added a toggle button for Light/Dark mode on the fjåge Web Shell. However, many modern web frameworks support dark mode, and it's common to embed the WebShell as an iframe inside web apps made from such a framework.

So we add two features to WebShell.

- WebShell now looks at the `class` attribute on the parent `HTML` element and if it finds a `light` or `dark` class attribute uses that as a hint to hide its toggle button and use the parent element's mode (light/dark) as it's own mode.
- WebShell also looks at an attribute on the `iframe` called `hideModeToggle` If that's set to true, then it also hides its toggle button.
- The WebShell accepts a message sent using the `window.postMessage` API with a `theme` property which defines the theme to be either `light` or `dark`. This allows the container web app to update the WebShell that the theme has changed.

Finally, we updated the button style for the toggle button.

### DIFF
--- a/src/main/resources/org/arl/fjage/web/shell/index.html
+++ b/src/main/resources/org/arl/fjage/web/shell/index.html
@@ -176,7 +176,7 @@
       if (hideModeToggle){
         darkmodeBtn.parentElement.style.display = 'none';
         window.addEventListener('message', evt => {
-          if (evt.data && evt.data.theme){q
+          if (evt.data && evt.data.theme){
             setTheme(term, evt.data.theme);
           }
         });
--- a/src/main/resources/org/arl/fjage/web/shell/index.html
+++ b/src/main/resources/org/arl/fjage/web/shell/index.html
@@ -73,18 +73,6 @@
     let delta = 200;
     let fitAddon;
     let currentTheme = null;
-    let hideModeToggle = false;
-    let parentTheme = null;
-    if (window.parent.document.documentElement.classList.contains('dark')){
-      parentTheme = 'dark';
-      currentTheme = parentTheme;
-      hideModeToggle = true;
-    }else if (window.parent.document.documentElement.classList.contains('light')){
-      parentTheme = 'light';
-      currentTheme = parentTheme;
-      hideModeToggle = true;
-    }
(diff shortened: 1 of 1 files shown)


src/main/resources/org/arl/fjage/web/shell/index.html
```diff
@@ -36,88 +45,50 @@
...
-    var delta = 200;
-    var fitAddon;
+    let defaultCursorWidth;
+    let rtime = 0;
+    let resizing = false;
+    let delta = 200;
+    let fitAddon;
+    let currentTheme = null;
+    let hideModeToggle = false;
+    let parentTheme = null;
```
ettersi, line 77: This is not really used anywhere. Delete?

src/main/resources/org/arl/fjage/web/shell/index.html
```diff
@@ -36,88 +45,50 @@
...
+    if (window.parent.document.documentElement.classList.contains('dark')){
```
ettersi, line 78: Move this to the `load` event listener so all the initialisation is in one place?

ettersi (approved):
LGTM other than a few tiny nitpicks 👍 

------

> * WebShell now looks at the `class` attribute on the parent `HTML` element and if it finds a `light` or `dark` class attribute uses that as a hint to hide its toggle button and use the parent element's mode (light/dark) as it's own mode.

*Technically* this could conflict with another component which uses toplevel `light` / `dark` class attributes for some other purpose. Super unlikely, of course, and probably not worth changing, but it does trigger some voices in my brain...

notthetup:
> Technically this could conflict with another component which uses toplevel light / dark class attributes for some other purpose. Super unlikely, of course, and probably not worth changing, but it does trigger some voices in my brain...

Indeed, there is no other way to do this other than assume that the parent will send the `postMessage` at initialization. There are lots of issues there as well with things being loaded asynchronously. 

On top of that adding a `class` to the `html` is pretty rare and I've only seen frameworks do it for this purpose. So I think we should be safe for now.

PR #312, opened 2024-06-29
Fix agentsForService test in test_fjage.c

brodiealexander:
Hi Fjåge team!

I'm opening this PR in the hopes of fixing the `agentsForService` test in `test_fjage.c`. In the current version, the test submits a null pointer in its call to fjage_agents_for_service in place of the allocated array of pointers it is supposed to submit. I've tested this version against UnetStack by changing the port to 1101 and changing the agent it looks for to be "websh", but I haven't been able to test against vanilla Fjage, as the quickstart script and build directions haven't been working for me. 

Best wishes, 
Brodie

### DIFF
--- a/src/main/java/org/arl/fjage/shell/ConsoleShell.java
+++ b/src/main/java/org/arl/fjage/shell/ConsoleShell.java
@@ -12,8 +12,10 @@ package org.arl.fjage.shell;
 
 import java.io.*;
 import java.util.logging.Logger;
+
 import org.arl.fjage.connectors.ConnectionListener;
 import org.arl.fjage.connectors.Connector;
+import org.arl.fjage.connectors.WebSocketConnector;
 import org.jline.reader.*;
 import org.jline.terminal.Terminal;
 import org.jline.terminal.TerminalBuilder;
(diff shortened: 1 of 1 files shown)


mchitre (commented):
Thanks @brodiealexander.

While the new implementation is more generic, it does not test anything differently from the old implementation, since `n = 1`. I don't see the point in making the test more complicated to test exactly what it was testing before. On the other hand, if we updated the test to have multiple shell agents and tested that we got multiple agents back, I would be happy to accept that change.

mchitre:
> As a side note: the `fjage_quickstart.sh` script began working for me when I swapped out gson-2.8.5 with gson-2.10.1

Thanks for this information. Would you mind opening a PR to update this (and perhaps other dependencies in the quick start script)?

brodiealexander:
> Thanks @brodiealexander.
> 
> While the new implementation is more generic, it does not test anything differently from the old implementation, since `n = 1`. I don't see the point in making the test more complicated to test exactly what it was testing before. On the other hand, if we updated the test to have multiple shell agents and tested that we got multiple agents back, I would be happy to accept that change.

I agree with your assessment here. I might look into this again in the future, but for now I'll close it.

Thanks!

PR #318, opened 2024-07-18
Allow the _to_json of AgentID to be used to generate the JSON for AgentIDs

notthetup:
In `Message` constructors of fjage.py we were forcing all the `AgentID` types to be converted to `v.name` before setting their values on the message objects. This looses information about if a `AgentID` is a `Topic` etc.

So instead, we let the `_to_json` method of `AgentID` do the proper stringification when the Message is converted to JSON.


mchitre:
Any idea why we had a special case handling for `AgentID` in the first place?

mchitre (approved):

PR #322, opened 2024-08-29
Log error thrown in parameter setter/getter at finest level

notthetup:
If an Agent parameter setter/getter method throws an exception, it's ignored by the `ParameterMessageBehavior`. But it's handy to log the exception so that it can be tracked. For now logging it at `FINE`.


mchitre (approved):
Good compromise to log at FINE level

PR #334, opened 2024-12-02
Make shell messages compliant with JSON

mchitre:
Shell messages (`ShellExecReq`, `GetFileReq`, `PutFileReq` and `GetFileRsp`) used private field names that did not agree with Javabean property names. This made the JSON fields be different from the Javabean properties. To avoid confusion with this, this PR unifies both. This is the convention we now follow with all messages.

Since this PR only modified internal field names, this is technically non-breaking. However, JS and C gateways/tests refer to private field names and so also had to be updated.


notthetup (approved):

notthetup:
This is likely to be a breaking change for the Gateway consumers most of them directly wrote to a field with the same name as the JSON element.

PR #337, opened 2024-12-19
fix(Behavior): Use atomic lookup instead of `synchronized` in `restart()`

ettersi:
Fixes a potential deadlock between the `Agent` and `Behavior` locks. X-ref https://github.com/org-arl/unet/issues/2057

Currently marked as "Draft" since we are still testing this. Will mark as ready and ask for review once the test was successful. 

Should be ported to `master` once merged.

### DIFF
--- a/src/main/java/org/arl/fjage/Behavior.java
+++ b/src/main/java/org/arl/fjage/Behavior.java
@@ -149,9 +149,13 @@ public class Behavior implements Comparable<Behavior> {
    *
    * @see #block()
    */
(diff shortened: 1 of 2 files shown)


notthetup:
While at it, would it be possible to add a test against this type of issues?

mchitre:
Concurrency bugs are difficult to test for, other than through stress/load tests. We do have some load tests in the current CI.

mchitre (approved):

PR #340, opened 2025-01-27
Seed the RNG used to generate UUID to reduce duplicate message IDs

notthetup:
`rand()` requires seeding to ensure that the UUIDs generated are random between multiple invocations of a program built with fjage C gateway.

### DIFF
--- a/gateways/c/src/fjage.c
+++ b/gateways/c/src/fjage.c
@@ -549,6 +549,7 @@ fjage_gw_t fjage_tcp_open(const char* hostname, int port) {
 #else
   fgw->intr = 0;
 #endif
+  srand(get_time_ms());
   char s[64];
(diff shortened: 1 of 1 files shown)


gateways/c/src/fjage.c
```diff
@@ -619,6 +621,7 @@ fjage_gw_t fjage_rs232_open(const char* devname, int baud, const char* settings)
 #else
   fgw->intr = 0;
 #endif
+  // seed random number generator
```
mchitre, line 624: What's this comment here for?

mchitre (approved):

gateways/c/src/fjage.c
notthetup: Ooops. Missing the `srand` here.

PR #344, opened 2025-03-12
Force WebServer to sync files after writing

notthetup:
When we upload a file using HTTP PUT multi-part through the WebServer.addupload functionality, we want to sync the file down to the filesystem before responding to ensure that file is safe.

This change does that.

### DIFF
--- a/src/main/java/org/arl/fjage/connectors/WebServer.java
+++ b/src/main/java/org/arl/fjage/connectors/WebServer.java
@@ -424,6 +424,13 @@ public class WebServer {
     public UploadHandler(MultipartConfigElement multipartConfig, Path outputDir) {
       super();
       this.multipartConfig = multipartConfig;
+      if (!Files.exists(outputDir)){
+        try {
+          Files.createDirectories(outputDir);
+        }catch (IOException ex){
+          log.warning("Unable to create directory : " + outputDir.toString());
+        }
+      }
       this.outputDir = outputDir;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/arl/fjage/connectors/WebServer.java
```diff
@@ -443,9 +450,18 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
           filename = URLEncoder.encode(filename, String.valueOf(StandardCharsets.UTF_8));
           Path outputFile = outputDir.resolve(filename);
           try (InputStream inputStream = part.getInputStream();
-               OutputStream outputStream = Files.newOutputStream(outputFile, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
+               FileOutputStream outputStream = new FileOutputStream(outputFile.toFile())) {
             IO.copy(inputStream, outputStream);
+            // Force data and metadata to disk
+            outputStream.getChannel().force(true);
+            // Ensure lowest-level disk synchronization
+            outputStream.getFD().sync();
```
ettersi, line 458: I believe the `outputStream.getChannel().force(true);` should be enough? From the [docs](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileChannel.html#force-boolean-):

> If this channel's file resides on a local storage device then when this method returns it is guaranteed that all changes made to the file since this channel was created, or since this method was last invoked, will have been written to that device. This is useful for ensuring that critical information is not lost in the event of a system crash.

src/main/java/org/arl/fjage/connectors/WebServer.java
```diff
@@ -443,9 +450,18 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
...
             out.printf("%s%n", outputFile);
+          }catch (Throwable ex){
+            log.warning("Unable to save file : " + filename + " : " + ex.toString());
```
ettersi, line 461: I know there's other lines analogous to this one in this file, but can we do
```suggestion
            log.log(Level.WARNING, "Unable to save file : " + filename, ex);
```
(requires importing `java.util.logging.Level`)? This prints the whole stack trace, while your version prints only the error message.

src/main/java/org/arl/fjage/connectors/WebServer.java
notthetup: > If the file does not reside on a local device then no such guarantee is made.

I am not sure what "local device" actually implies here. Would a disk attached over USB be considered "local"?

src/main/java/org/arl/fjage/connectors/WebServer.java
notthetup: Right, we probably want to go about and do this in most other places where we log an exception.

ettersi (approved):

PR #346, opened 2025-04-11
Update `getName` for SerialPortConnector

notthetup:
Moving from using `jserial`'s `com.getDescriptivePortName` to using the `devname` is provided in the constructor to generate the `getName`, to have a more consistent name for a `SerialPortConnector` in the format `serial://<devname>:<baud>,<settings>`


mchitre (approved):

notthetup:
Reverted this since it broke the build. Will revisit it in another way.

PR #347, opened 2025-04-16
Non aggregating WebSocketConnector

notthetup:
We've had a `WebSocketConnector` which was actually a `WebSocketHubConnector` (similar to `TcpHubConnector`) since it aggregated all the input and output streams. 

So I've renamed the `WebSocketConnector` to `WebSocketHubConnector` and create a new `WebSocketServer` which creates a `WebSocketConnector` when a new connection comes in, and that is bound managed as a "slave" by the `MasterContainer`, similar to how a `TcpServer` and `TcpConnector` work.

To expose this functionality, on the `MasterContainer` we now have `container.openWebSocketServer` and `container.closeWebSocketServer` methods to be able to enable these servers.

To support a non-aggregating WebSocketConnector in `fjage.js` we also enable sending out `WANTS_MESSAGES_FOR` messages.

This is potentially a breaking change, since we renamed the class. I would propose changing the MAJOR version change with this.


src/main/java/org/arl/fjage/remote/MasterContainer.java
```diff
@@ -384,9 +385,9 @@ public void shutdown() {
       slaves.clear();
       needsCleanup = false;
     }
-    if (listener != null) {
-      listener.close();
-      listener = null;
+    if (tcpListener != null) {
+      tcpListener.close();
+      tcpListener = null;
     }
```
ettersi, line 391: We should close the `websocketListener` also, no?

notthetup: Good catch!

ettersi (approved):

src/main/java/org/arl/fjage/connectors/WebSocketConnector.java
```diff
@@ -1,266 +1,168 @@
-/******************************************************************************
-
-Copyright (c) 2018, Mandar Chitre
-
-This file is part of fjage which is released under Simplified BSD License.
-See file LICENSE.txt or go to http://www.opensource.org/licenses/BSD-3-Clause
-for full license details.
-
-******************************************************************************/
-
```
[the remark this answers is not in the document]
mchitre: OK with removing all licenses from the files if we wish to

mchitre (commented):
Do we really want this to be a breaking change? While I agree that a name change is cleaner, does it warrant a breaking release? Or could we have an additional constructor with a `hub` argument that defaults to `true`?

notthetup:
If we want to keep the backward compatibility, then I would suggest naming the new connector as `WebSocketDiscreteConnector` to indicate that it's not a "Hub" connector.

But I think in the longer run, we'll be better off with the breaking change and a Major version change.

notthetup:
Adding changes to ChangeLog and Readme to explain the deprecation of `1.15.x` releases and revving the Major version to `2.0.0` for both fjåge and fjage.js.

mchitre (approved):

notthetup:
Just to note, I have removed all the changes about v2.0.0. We'll add those in a separate PR.

PR #359, opened 2025-07-28
Add `AutosuggestionWidgets` and `DefaultHistory` to ConsoleShell

notthetup:
The `ConsoleShell` which is used for both the Terminal based fjage shell and also the web based fjage shell uses [Jline](https://jline.org/) to process the inputs.

Modern JLine supports [Autosuggestion](https://jline.org/docs/advanced/autosuggestions) and [History](https://jline.org/docs/history).

This PR enables those two features in fjage.

## History

Reverse search of typed commands was always supported, but this PR adds persistent history which is saved in a file on disk. This can be enabled by passing a the file `Path` to the `ConsoleShell` constructor.

The default run script in `initrc.groovy` implements, but setting a `.fjage-shell-history` file that is stored in the working directory.

The default size of history is set to 1000 items and it can be cleared by deleting the `.fjage-shell-history` file.

## Autosuggestion

The AutoSuggestion widget will suggest commands from the history, which start with the string that the user has started typing.

---

Both these features are turned off for ConsoleShell that connect to `DUMB` Terminals on local systems, and or where there is no `ScriptEngine` set on the `ConsoleShell`.

P.S. This PR requires https://github.com/org-arl/fjage/pull/358 to be merged.

### DIFF
--- a/src/main/java/org/arl/fjage/shell/ConsoleShell.java
+++ b/src/main/java/org/arl/fjage/shell/ConsoleShell.java
@@ -1,16 +1,16 @@
 /******************************************************************************
 
-Copyright (c) 2018-2019, Mandar Chitre
+ Copyright (c) 2018-2019, Mandar Chitre
 
-This file is part of fjage which is released under Simplified BSD License.
-See file LICENSE.txt or go to http://www.opensource.org/licenses/BSD-3-Clause
-for full license details.
-
-******************************************************************************/
+ This file is part of fjage which is released under Simplified BSD License.
+ See file LICENSE.txt or go to http://www.opensource.org/licenses/BSD-3-Clause
+ for full license details.
+ ******************************************************************************/
 
 package org.arl.fjage.shell;
 
 import java.io.*;
+import java.nio.file.Path;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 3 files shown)


ettersi:
I believe when we have stacked PRs like this one and #358, it's better to ask for the second one to be merged into the first one instead of `master`. This way, the second PR will show only the changes in addition to the first PR, and if I'm not wrong then GitHub will automatically update the merge target for the second PR to `master` once the first PR is merged.

src/main/java/org/arl/fjage/shell/ConsoleShell.java
```diff
@@ -123,25 +157,15 @@ public void init(ScriptEngine engine) {
...
-          return parse(s, cursor);
-        }
-        @Override
-        public boolean isEscapeChar(char ch) {
-          return false;
-        }
-      };
-      console = LineReaderBuilder.builder().parser(parser).terminal(term).build();
-      console.setVariable(LineReader.DISABLE_COMPLETION, true);
-      console.setOpt(LineReader.Option.ERASE_LINE_ON_FINISH);
```
ettersi, line 143: Was this removal done on purpose? It changes the shell behaviour quite significantly. 

On master, when entering `[<enter>` I get
```
> [
- 
```

On this branch, I get
```
> [
[: ']' expected
```
and then the fjage shell gets closed!

notthetup: The custom parser was causing NPE in the autosuggestion widget. The `setVariable` and `setOpt` changes were from the TabCompleter tests I did. I can revert them back.


Will also fix this issue

notthetup: OK. I tried this. I can't seem to reproduce this. Anyway I added back the `setVariable` and `setOpt`.

https://asciinema.org/a/HbablmoZ3PerEnsSCw0Tfp8SF

ettersi (approved):

PR #363, opened 2025-07-28
Improving fjåge.js timeouts

notthetup:
This pull request improves timeouts available of various Gateway and AgentId methods in fjage.js.

### Gateway Improvements

- Previously the Gateway constructor had an option argument `new Gateway({timeout: 1000})` which set the default timeout for all requests including the fjage container level queries like `agentForService`, `agents`, etc. This is still supported, but we now add the ability to set a timeout for each request individually using an argument to `agentForService`, `agentsForService`, `agents` and `containsAgent` methods. 
- The default timeout remain the same (the default is 8 times the Gateway's timeout, which itself was 1000ms by default).
- With the new argument, you can now specify a different timeout for each request, e.g. `gw.agents(10000)` will set the timeout for that request to 10 seconds instead of the default 8 seconds. You can even set it to -1 to disable the timeout for that request, e.g. `gw.agents(-1)` will wait indefinitely for the response.

### AgentID Improvements

- We also default the timeout for all requests sent to an Agent using the `AgentId` methods (`aid.get()`, `aid.set()`, `aid.request()`, etc.) to default to the Gateway's timeout. Previously, this was set to 1000ms for `request` and 5000ms for `set` and `get`. By copying from the Gateway's timeout and multiplying it accordingly to the previous values, we keep the default behavior consistent.
- We now add the ability to change the default timeout for any requests sent to a specific Agent by adding a `setTimeout` method to the AgentId class. This allows you to set a default timeout for all requests sent to that AgentId.

### DIFF
--- a/gateways/js/src/agentid.js
+++ b/gateways/js/src/agentid.js
@@ -3,6 +3,9 @@ import { ParameterReq  } from './message.js';
 import { Gateway } from './gateway.js';  // import Gateway class for type checking. Remove if not needed.
 import { Message } from './message.js';  // import Message class for type checking. Remove if not needed.
 
+const DEFAULT_TIMEOUT = 10000; // Default timeout for non-owned AgentIDs
+
+
 /**
 * An identifier for an agent or a topic. This can be to send, receive messages, and set or get parameters
 * on an agent or topic on the fjåge master container.
@@ -18,6 +21,7 @@ export class AgentID {
     this.name = name;
     this.topic = topic;
     this.owner = owner;
+    this._timeout = owner ? owner._timeout : DEFAULT_TIMEOUT; // Default timeout if owner is not provided
   }
 
   /**
@@ -54,10 +58,10 @@ export class AgentID {
   * Sends a request to the agent represented by this id and waits for a reponse.
   *
   * @param {Message} msg - request to send
-  * @param {number} [timeout=1000] - timeout in milliseconds
+  * @param {number} [timeout=owner.timeout] - timeout in milliseconds
   * @returns {Promise<Message>} - response
   */
-  async request(msg, timeout=1000) {
+  async request(msg, timeout=this._timeout) {
     msg.recipient = this;
     if (this.owner) return this.owner.request(msg, timeout);
     else throw new Error('Unowned AgentID cannot send messages');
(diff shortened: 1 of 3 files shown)


notthetup:
I removed all the bits about metrics. Just the changes to timeouts now.

ettersi:
> > It is already quite difficult to keep track of which timeout is being used if I don't specify one myself. This PR is going to make this problem even worse. That's not really an objection to the specific changes here, just voicing out some bigger-picture concerns I have.
> 
> The idea here is to allow the user of fjage.js to specific the timeout instead using some magic numbers internally. If anything I would suspect this would help make things more clear.

My concern is a scenario where we get some logs which indicate that a request timed out, and now we have to go figure out what timeout was used for this request. This would be straightforward if we used a single, global, immutable default timeout but can be tricky once you have gateway-level and AgentID-level timeout overrides that can be changed by anyone who has access to the respective objects.

gateways/js/src/agentid.js
```diff
@@ -106,10 +117,10 @@ export class AgentID {
   * @param {(string|string[])} params - parameters name(s) to be set
   * @param {(Object|Object[])} values - parameters value(s) to be set
   * @param {number} [index=-1] - index of parameter(s) to be set
-  * @param {number} [timeout=5000] - timeout for the response
+  * @param {number} [timeout=this._timeout*5] - timeout for the response
   * @returns {Promise<(Object|Object[])>} - a promise which returns the new value(s) of the parameters
   */
-  async set (params, values, index=-1, timeout=5000) {
+  async set (params, values, index=-1, timeout=this._timeout*5) {
```
ettersi, line 123: Why the (arbitrary?) factor five here?

gateways/js/src/gateway.js
```diff
@@ -516,11 +520,12 @@ export class Gateway {
 
   /**
   * Gets a list of all agents in the container.
+  * @param {number} [timeout=this._timeout*8] - timeout in milliseconds
   * @returns {Promise<AgentID[]>} - a promise which returns an array of all agent ids when resolved
   */
-  async agents() {
+  async agents(timeout=this._timeout*8) {
```
ettersi, line 526: As above. Why have seemingly arbitrary multiplication factors in some places?

notthetup:
> Timeouts are anyway supposed to be a last-resort fix to prevent the system from deadlocking, no? So maybe the solution is to just use the max of all the individual timeouts everywhere?

Good point. I was looking at reducing them, but all that does is reduces latency of catching and dealing with errors. So instead we can set the default to something like 10000ms and then we don't have to do any of these multipliers etc.

I'll update the PR accordingly.

---

> Also, when you want to change a parameter somewhere, explicit parameters force you to go and propagate that change throughout the entire stack, which seems annoyingly tedious at first but is actually a good thing because it requires you to explicitly go through everything that could potentially have been broken by that change.

The issue here is always forgetting it in one place.

But let's leave this change aside for now. We can revisit it later if we need it.

ettersi:
> > Also, when you want to change a parameter somewhere, explicit parameters force you to go and propagate that change throughout the entire stack, which seems annoyingly tedious at first but is actually a good thing because it requires you to explicitly go through everything that could potentially have been broken by that change.
> 
> The issue here is always forgetting it in one place.

Yeah, to optimise for this particularly purpose the timeout argument would have to be mandatory. But that's of course breaking and annoying, so definitely not proposing this seriously. This whole discussion is just a tangent to this PR anyway, so happy to leave this for now as you suggested.

PR #364, opened 2025-07-28
chore(fjagejs): refactoring to split various functionality into indiv…

notthetup:
…idual files


gateways/js/test/spec/fjage.spec.js
```diff
@@ -335,6 +335,56 @@ describe('A Gateway', function () {
     gw.close();
   });
 
+  it('should be able to request for all agents from the master container', async function() {
+    const gw = new Gateway(gwOpts);
+    await delay(300);
```
ettersi, line 340: Why are we waiting here? Opening a gateway and immediately start fetching data is much more common, no?

ettersi: I'm also not sure why we're having a close delay in most tests, but I'm happy to punt that for another day.

ettersi (approved):

ettersi:
Saw your most recent commit, still looks good to me 👍

PR #370, opened 2025-08-20
Use UUID7 

notthetup:
We need UUID when creating new `Message.msgID` and `JSONMessage.id` for fjage. Historically we used various UUID/GUID implementations mainly from standard library but sometimes custom. The main requirement is uniqueness across all messages send in a single fjage "universe".

While the previous implementations of UUID/GUID were sufficient for that, the new (and upcoming) UUID7 standard has a couple of useful features. 

- Since the MSB of the UUID7 is a millisecond level timestamp, it's easy to sort messages using the `msgID`.
- Since the MSB of the UUID7 is a millisecond level timestamp, we can use the timestamp to know when messages were created. This could be useful for collecting metrics and logging how long messages stayed in queues etc. 

See more : http://uuid7.com/

### DIFF
--- a/gateways/c/src/fjage.c
+++ b/gateways/c/src/fjage.c
@@ -74,6 +74,7 @@ for full license details.
 static fjage_msg_t fjage_msg_from_json(const char* json);
 static void fjage_msg_write_json(fjage_gw_t gw, fjage_msg_t msg);
 static void fjage_msg_set_sender(fjage_msg_t msg, fjage_aid_t aid);
+static long get_time_ms(void);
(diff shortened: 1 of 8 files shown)


mchitre (approved):
Have we benchmarked the performance of UUID7 vs the Java UUID implementation we were using to know the impact? Also, we need to add this to `Fjage.jl` too.

notthetup:
> Have we benchmarked the performance of UUID7 vs the Java UUID implementation we were using to know the impact? Also, we need to add this to `Fjage.jl` too.

No. I can do this.

PR #372, opened 2025-09-10
fjage.py v2.0.0

notthetup:
# Re-implementation of Python fjåge Gateway

The Python fjåge Gateway had diverged significantly from the Gateway specification over the years and lacked many of the improvements present in the JavaScript and Java implementations.

This PR **re-implements the Python Gateway from scratch**, aligning it with the architecture of the JavaScript Gateway. The code is now modular, with each class in its own file, and the `JSONMessage` class extracted for clarity and reuse.

---

## Key Improvements

* **Modern packaging**
  Migrated to `pyproject.toml` for project metadata and build configuration.

* **Connector abstraction**
  Introduced a `Connector` ABC, with `TCPConnector` as the default implementation. This enables plugging in custom connectors (e.g., Serial).

* **Improved thread safety**
  Made all the internal queues and dictionaries of pending messages and actions thread safe between the user and the background TCP reader thread.	

* **Testing against a live container**
  Added unit tests for core features (`AgentID`, `Gateway`, `Message`) using a Java Master Container, mirroring the JS test strategy.

* **Better developer ergonomics**

  * Docstrings and type hints across most APIs
  * Context manager support: `with Gateway('localhost', 5081) as gw:`
  * Cleaner logging (no spurious messages, lowered default levels)
  * CI pipeline updated with modern workflows

* **Documentation**
  Automated API docs generation, published at: [https://org-arl.github.io/fjage/pydoc/](https://org-arl.github.io/fjage/pydoc/)

* **New API features**

  * `AgentID.get()` now returns *all* parameters of an agent as a `dict`.

---

## Major Changes

* **Pythonic naming**

  * `Gateway.isConnected()` deprecated for → `Gateway.is_connected()`
  * `Gateway.getAgentID()` deprecated for → `Gateway.agent_id()`
  * `Gateway.containAgent()` now has an alias `Gatewa.contains_agent()`
  * `Gateway.agentForService()` now has an alias `Gateway.agent_for_service()`
  * `Gateway.agentsForService()` now has an alias `Gateway.agents_for_service()`

* **Numpy dependency removed**

  * Messages previously using `__isComplex` were returned as numpy arrays.
  * Now returned as standard Python lists of complex numbers.
  * If a numpy array is passed as a message attribute, it is converted to a list via dynamic import.


notthetup:
> Agreed to make `numpy` an injected dependency. The user can pass `numpy` to the Gateway constructor, in which case, we will use numpy to parse/convert the arrays.

After further analysis, we agree that we can do this as a performance enhancement if required. We won't target this in v2.0.0.

notthetup:
> Correct. They're used by sphinx/pydoc to generate the API docs.
> 
> >

Yup. Removed `docs/pydocs/.doctrees` and added to .gitignore.

mchitre (approved):

PR #374, opened 2025-09-29
fix: mjs mapping

ymtoo:
This PR initializes `resHandler` MIME types if `resHandler.getMimeTypes()` is `null`.


ettersi:
I suspect this PR will not receive much attention unless we ask someone for review. Hence requesting a review from @notthetup who is probably in the best position to assess whether this is the right place to make the necessary changes.

ymtoo:
@notthetup Could you help to merge the changes into the `webserver-mjs` branch?

PR #398, opened 2026-04-10
Fix remote container correctness and websocket write handling

notthetup:
Improve MasterContainer/WebSockets correctness and thread safety.

- Hardened remote container concurrency and shutdown paths across MasterContainer, SlaveContainer, ConnectionHandler, JsonMessage, MessageQueue, Agent, and Container, including the async/query handling refactor.
- Fixed stream lifecycle in the connector layer: BlockingByteQueue now has explicit close semantics, and PseudoInputStream/PseudoOutputStream use that closed state instead of nulling the queue.
- Added regression coverage for blocked readers/unblocking-on-close behavior.


notthetup:
Here is a version to test. It should be a drop in replacement for fjage 2.4.0

[fjage-2.4.0.jar.zip](https://github.com/user-attachments/files/27466024/fjage-2.4.0.jar.zip)

mchitre:
Many of the comments from Copilot are valid. Suggest addressing them before I review in detail.

src/main/java/org/arl/fjage/remote/MasterContainer.java
```diff
@@ -462,11 +417,99 @@ private void openTcpServer(int port) {
...
+      futures.add(future);
+      future.thenAccept(rsp -> {
+        if (predicate.test(rsp)) result.complete(rsp);
+      });
+      future.whenComplete((rsp, ex) -> {
+        if (remaining.decrementAndGet() == 0) result.complete(null);
+      });
+    }
+    T match = result.join();
+    return Optional.ofNullable(match);
```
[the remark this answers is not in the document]
notthetup: I don't think this is a valid issue. In fact this the optimization. `firstMatch` runs the `fn` on each of the `handlers` and uses the `predicate` on the return value of the `fn` to decide if it was success.

It's used here

https://github.com/org-arl/fjage/blob/2ab9e22383c24b8b245c64573cf2282d90e9ecf2/src/main/java/org/arl/fjage/remote/MasterContainer.java#L209-L216

and

https://github.com/org-arl/fjage/blob/2ab9e22383c24b8b245c64573cf2282d90e9ecf2/src/main/java/org/arl/fjage/remote/MasterContainer.java#L270-L280

In both cases it will send out the requests in parallel which is what we want.


---

>  trigger request timeouts that mark unrelated slow-but-healthy connections as dead/closed

I don't see why this would happen.

@mchitre Do you see why this would happen?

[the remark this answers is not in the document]
mchitre: It is true that this will send to all slaves without waiting for answers, and so will cause more traffic. But the traffic will be in parallel, and once a positive response is obtained, the other responses will just be ignored. Network will be loaded slightly more, but clock time performance might be better. I have no objections to this approach.

src/main/java/org/arl/fjage/connectors/BlockingByteQueue.java
mchitre, line 162: Should we return `-1` if the queue is closed? Currently returns `0`, while `read()` returns `-1`.

src/main/java/org/arl/fjage/remote/ConnectionHandler.java
```diff
@@ -235,33 +229,30 @@ synchronized void println(String s) {
     }
   }
 
-  void printlnQueued(String s) {
```
mchitre, line 238: With this gone, a blocked `send()` on one slave blocks all slaves. For TCP connections, this happens when a send buffer fills. Tested it with a stalled TCP slave and all slaves get blocked.

Also, because this is synchronized, the ALIVE watchdog can't close the connection – I think this is one of the issues Copilot commented on as well.

Directory operations can also block on a bad TCP connection write in the same way.

src/main/java/org/arl/fjage/remote/ConnectionHandler.java
```diff
@@ -32,28 +32,18 @@ public class ConnectionHandler extends Thread {
...
-  private final Map<String,Object> pending = Collections.synchronizedMap(new HashMap<>());
+  private final ConcurrentMap<String,Object> pending = new ConcurrentHashMap<>();
   private final Deque<String> failed = new ArrayDeque<>(FAILED_SIZE);
   private final Logger log = Logger.getLogger(getClass().getName());
   private final RemoteContainer container;
-  private boolean alive;
+  private volatile boolean alive;
   private final boolean keepAlive;
   private final boolean closeOnDead;
   private final ExecutorService pool = Executors.newSingleThreadExecutor();
```
mchitre, line 42: Recommend setting `setRemoveOnCancelPolicy(true)` on pool to avoid accumulation of cancelled tasks to be cleaned after a default timeout of 30 seconds.

src/main/java/org/arl/fjage/Container.java
```diff
@@ -50,9 +50,9 @@ public class Container {
   protected Map<AgentID,Set<Agent>> topics = new HashMap<>();
   protected Map<String,Set<AgentID>> services = new HashMap<>();
   protected Logger log = Logger.getLogger(getClass().getName());
-  protected boolean running = false;
+  protected volatile boolean running = false;
   protected boolean initing = false;
```
mchitre, line 54: needs `volatile`?

src/main/java/org/arl/fjage/remote/MasterContainer.java
```diff
@@ -354,24 +314,23 @@ AgentID[] localAgentsForService(String service) {
...
-      }
-    }
+    for (ConnectionHandler slave: slaves)
+      if (!slave.isAlive()) slave.start();
     if (!slaves.isEmpty()) {
       log.fine("Waiting for slaves...");
       boolean allAlive = false;
-      for (int i = 0; !allAlive && i < TIMEOUT/100; i++) {
+      // Bound startup sync to a short local wait; request timeouts are handled later.
+      for (int i = 0; !allAlive && i < 50; i++) {
```
mchitre: Use a defined constant instead of magic number `50`?

mchitre:
Also, ran a bunch of tests on this PR. Genuine improvements in performance. The tests involved several TCP and web clients connected to a master container exchanging messages at high load, and having some of them misbehave or die and looking at effect on performance.

Under combined message + directory-query load the master branch collapses (406 msg/s, undelivered messages after 60 s) while the PR branch sustains 37–68k msg/s with zero loss and zero duplicates. Web socket throughput improves ~200×. Reconnect, malformed-JSON, and shutdown-race behavior are all clean.

One finding: message relay writes now happen synchronously on the sending agent’s thread, so a slow or dead-but-connected peer can block master agents (with watchdog unable to recover as noted in [review comments](https://github.com/org-arl/fjage/pull/398#discussion_r3592732878)).

Tabulated result summary below:
image

T3c failure is the finding above. T7 failure is https://github.com/org-arl/fjage/pull/398#discussion_r3592706007

src/main/java/org/arl/fjage/remote/ConnectionHandler.java
notthetup: I don't think we cancel any tasks from `pool`.

src/main/java/org/arl/fjage/Container.java
notthetup: It's read from the various Agent's threads. Having it `volatile` will ensure that Agents don't see different state. Especially during startup/shutdown when this changes.

src/main/java/org/arl/fjage/remote/ConnectionHandler.java
notthetup: Ah yes. So this was one of the changes that made sense when I added the async sending. Because the PR was getting too large, I split the change into https://github.com/org-arl/fjage/pull/422.

So I'll move this change to that PR instead and revert to the older behaviour.

notthetup:
> Also, ran a bunch of tests on this PR. Genuine improvements in performance. The tests involved several TCP and web clients connected to a master container exchanging messages at high load, and having some of them misbehave or die and looking at effect on performance.
> 
> Under combined message + directory-query load the master branch collapses (406 msg/s, undelivered messages after 60 s) while the PR branch sustains 37–68k msg/s with zero loss and zero duplicates. Web socket throughput improves ~200×. Reconnect, malformed-JSON, and shutdown-race behavior are all clean.
> 
> One finding: message relay writes now happen synchronously on the sending agent’s thread, so a slow or dead-but-connected peer can block master agents (with watchdog unable to recover as noted in [review comments](https://github.com/org-arl/fjage/pull/398#discussion_r3592732878)).
> 
> [snip]
> 
> T3c failure is the finding above. T7 failure is [(earlier comment)](https://github.com/org-arl/fjage/pull/398#discussion_r3592706007)

Would it be possible to commit this test as standard Fjage tests? I think it could be useful to track some of these metrics.

src/main/java/org/arl/fjage/Container.java
mchitre: The question was for `initing` – should we be making it `volatile`, since you felt the need to make `running` and `inited` `volatile`.

notthetup: `initing` doesn't seem to be used by other threads. So no need.

src/main/java/org/arl/fjage/remote/ConnectionHandler.java
```diff
@@ -272,10 +263,11 @@ JsonMessage printlnAndGetResponse(String s, String id, long timeout) {
 
   synchronized void close() {
     if (conn == null) return;
-    if (keepAlive && container instanceof SlaveContainer) println(SIGN_OFF);
+    if (keepAlive && container instanceof SlaveContainer) send(SIGN_OFF);
     conn.close();
     conn = null;
     out = null;
+    pending.clear();
     container.connectionClosed(this);
```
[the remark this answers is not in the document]
mchitre: @notthetup, still to resolve? Hold review until this is done?

src/main/java/org/arl/fjage/remote/SlaveContainer.java
```diff
@@ -118,12 +119,9 @@ public SlaveContainer(Platform platform, String name, String devname, int baud,
...
   public boolean authenticate(String creds) {
     if (master == null) return false;
-    JsonMessage rq = new JsonMessage();
-    rq.action = Action.AUTH;
+    JsonMessage rq = JsonMessage.createActionRequest(Action.AUTH);
     rq.creds = creds;
-    rq.id = UUID.randomUUID().toString();
-    String json = rq.toJson();
-    JsonMessage rsp = master.printlnAndGetResponse(json, rq.id, TIMEOUT);
+    JsonMessage rsp = master.request(rq, TIMEOUT);
```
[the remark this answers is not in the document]
mchitre: @notthetup this is not really addressed. The `master` null check above could pass and this could still NPE since `master` is `volatile` and cannot be cached by compiler. This is likely not a problem in practice, but just for completeness, shall we capture `master` in all methods before we do the check in a local variable and use that?

mchitre:
@notthetup couple of comments above to resolve before we are ready to merge.

src/main/java/org/arl/fjage/remote/MasterContainer.java
notthetup: Fixed.

mchitre:
I pushed a multi-container stress harness to this branch (f462d33a) to validate the concurrency changes: 1 master + 3 slave containers in one JVM over real localhost TCP/WS, with fault injection (slow peer, unresponsive/malformed peer, abrupt disconnects + flapping, mid-load shutdowns, WS storm, stream-close races) and loss/dup/latency/leak checks. It is **not** run by `gradlew test` or CI — run it on demand:

```sh
./gradlew loadTest                                              # all scenarios, ~6 min
./gradlew loadTest --tests 'org.arl.fjage.loadtest.T1LoadTest'  # one scenario
```

### Results on 3e9fea10: 8/10 pass

| Test | Result | Notes |
|------|--------|-------|
| T1 load (4 containers, ~20k msgs + concurrent directory ops) | ✅ | zero loss, zero dups (master branch fails this: throughput collapses to ~400 msg/s with undelivered messages) |
| T2 slow slave (10 s TCP backpressure) | ✅ | full recovery, no head-of-line blocking to healthy slaves |
| T3 unresponsive / malformed / silent peers | ✅ ×3 | dead connections pruned, relay unaffected |
| T4 abrupt loss + reconnect + 8× flapping | ✅ | zero loss on healthy slaves, no handler/thread leaks |
| T5 shutdown races (12 cycles mid-load) | ✅ | clean teardown every cycle |
| T6 WebSocket storm + abrupt disconnect | ✅ | ~17,600 msg/s vs ~83 msg/s on master (removed 10 ms/line throttle) |
| T7 stream close stress | ❌ ×2 | see below |

The `sendQueued()` relay added in 3e9fea10 works well: in T2 the worst agent-thread block during the stall dropped from **6.8 s to 3.9 ms**, and the T3c scenario (a client that connects but never reads — previously wedged the relaying agent thread permanently, with the ALIVE watchdog unable to close the connection) now runs clean. Ordering is preserved via the handler's single-thread pool.

### Remaining bug: `BlockingByteQueue.read(byte[])` close race

The `if (closed) return -1;` guard only covers a queue already closed at entry. If `close()` lands while `read(byte[])` is in progress, the inner `read()` returns −1 and the loop does `return i` with `i == 0` — so the caller gets 0 instead of −1, and anything looping on `>= 0` busy-spins. T7 reproduces this within a few rounds. Suggested fix inside the loop:

```java
int c = read();
if (c < 0) return i > 0 ? i : -1;
```

### Minor observations (not regressions from this PR)

- Directory-op requests (`getAgents()` etc.) still block ~9 s when a slave's socket is backed up — same on master; the request path does a synchronous write with no timeout.
- With the queued relay, a peer that never reads now grows the handler pool's unbounded queue instead of blocking the agent (the same trade-off the old `printlnQueued` had). The 60 s ALIVE watchdog bounds this for keep-alive connections; may be worth a bounded queue eventually.

T7's two failures are committed intentionally as reproductions — they should go green once the close race is fixed.

mchitre:
Tested the above one-liner fix and tests pass.

notthetup:
`org.arl.fjage.loadtest.T7StreamStressTest.queueCloseStress` seemed to fail sporadically, seemed like an OOM. I added an upper bound to how much the writers write in the test in https://github.com/org-arl/fjage/commit/698a21e16fcd3209f85e00a0638ac30fcb127689

That seems to fix it.

src/main/java/org/arl/fjage/remote/SlaveContainer.java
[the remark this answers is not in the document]
notthetup: OK. Fixed.

src/main/java/org/arl/fjage/remote/ConnectionHandler.java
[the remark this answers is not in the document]
notthetup: Done.

mchitre (approved):

PR #399, opened 2026-04-15
`@message` decorator in fjagepy to register new Message subtypes

notthetup:
Currently we use the `MessageClass` function in most Gateways (including fjagepy) to register new `Message` subtypes to that the Gateway can inflate them.

The `MessageClass` just generates a generic class and doesn't really allow definition of the fields of the type or shape of the Message subtype.

This PR introduces a `@message` decorator which can be used to register any standard Python class (with type hints and all) with fjagepy, such that it will be inflated properly if the Gateway receives a message of that type from the `MasterContainer`.

This is similar to the `@message` macro in https://github.com/org-arl/Fjage.jl

```python
from fjagepy import Message, message


@message('org.example.CustomReq')
class CustomReq(Message):
    def __init__(self, text=None):
        super().__init__()
        self.text:Optional[str] = text


outgoing = CustomReq('hello')
payload = outgoing.to_json()

# Later, when a message with the same clazz arrives from fjage:
incoming = Message.from_json(payload)
assert isinstance(incoming, CustomReq)
assert incoming.text == 'hello'
```

Also this fixes an issue to make fjagepy deal with keyword-safe fields (`from_`) consistently across assignment, serialization, and deserialization.

Also this PR adds `mypy` based type checking for the whole package and resolves some of the typechecking issues.


gateways/python/fjagepy/Message.py
```diff
@@ -162,24 +237,37 @@ def __init__(self, **kwargs):
...
-GetFileReq = MessageClass('org.arl.fjage.shell.GetFileReq')
-ShellExecReq = MessageClass('org.arl.fjage.shell.ShellExecReq')
-GetFileRsp = MessageClass('org.arl.fjage.shell.GetFileRsp')
+    if isinstance(arg, type):
+        return decorate(arg)
+
+    def decorator(class_: Type["Message"]) -> Type["Message"]:
+        return decorate(class_, arg)
+
+    return decorator
```
dixslyf: The type for `message()` is wrong. I believe this should be the right type, which makes `pylsp` happy. It's a little confusing because `message()` is either the decorator itself or a function returning the decorator.

 I've also replaced `Type` with `type` as that's the recommended way to do it for Python 3.9+.

```suggestion
@overload
def message(arg: type[Message]) -> type[Message]:
    ...

@overload
def message(arg: str) -> Callable[[type[Message]], type[Message]]:
    ...

def message(arg: type[Message] | str) -> Union[type[Message], Callable[[type[Message]], type[Message]]]:
    """Decorator to register a Message subclass for JSON inflation.

    Can be used as ``@message`` or ``@message('org.example.MyMessage')``.
    """

    def decorate(class_: type["Message"], fqcn: Optional[str] = None) -> type["Message"]:
        if not issubclass(class_, Message):
            raise TypeError('@message can only be used with Message subclasses')

        clazz_name = fqcn or class_.__dict__.get('__clazz__') or class_.__name__
        class_.__clazz__ = clazz_name
        return _register_message_class(class_, clazz_name)

    if isinstance(arg, type):
        return decorate(arg)

    def decorator(class_: type["Message"]) -> type["Message"]:
        return decorate(class_, arg)

    return decorator
```

gateways/python/fjagepy/Message.py
dixslyf:
```suggestion
    def parameters(self) -> dict[str, Any]:
```

dixslyf: Not sure why GitHub's UI is showing the diff incorrectly, at least on my end. This is the right diff:

```diff
-    def parameters(self) -> dict[str, any]:
+    def parameters(self) -> dict[str, Any]:
```

gateways/python/fjagepy/Message.py
```diff
@@ -1,21 +1,75 @@
...
+        return instance
+
 # Attempt to import numpy for array handling
 # If numpy is not available, we don't need to
 # handle numpy arrays in Message
 try:
     import numpy
 
-    def _serialize_numpy_array(value: numpy.ndarray, key: str, props: Dict) -> list:
+    def _serialize_numpy_array(value: Any, key: str, props: Dict) -> list:
```
dixslyf: Is there a reason for changing the type to `Any`?

gateways/python/fjagepy/Message.py
```diff
@@ -1,21 +1,75 @@
...
+    _MESSAGE_REGISTRY[class_.__name__] = class_
+
+    clazz_name = fqcn or getattr(class_, '__clazz__', None)
+    if isinstance(clazz_name, str) and clazz_name:
+        _MESSAGE_REGISTRY[clazz_name] = class_
+        _MESSAGE_REGISTRY[clazz_name.split('.')[-1]] = class_
+
+    return class_
+
+def _instantiate_message(class_: Type["Message"]) -> "Message":
```
dixslyf:
```suggestion
def _instantiate_message(class_: type["Message"]) -> "Message":
```

gateways/python/fjagepy/Message.py
```diff
@@ -1,21 +1,75 @@
...
 from .Performative import Performative
 from .Utils import UUID7
 
+if TYPE_CHECKING:
+    from .Gateway import Gateway
+
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
+_MESSAGE_REGISTRY: Dict[str, Type["Message"]] = {}
```
dixslyf: Python 3.9+ recommendation is to use `type` instead of `Type`.
```suggestion
_MESSAGE_REGISTRY: Dict[str, type["Message"]] = {}
```

gateways/python/fjagepy/Message.py
```diff
@@ -1,21 +1,75 @@
...
+def _normalize_field_name(name: str) -> str:
+    """
+    Normalize a field name by stripping trailing underscores from Python keywords.
+    """
+    if not name.startswith('_') and name.endswith('_') and keyword.iskeyword(name[:-1]):
+        return name[:-1]
+    return name
+
+
+def _message_clazz(class_: Type["Message"]) -> str:
```
dixslyf:
```suggestion
def _message_clazz(class_: type["Message"]) -> str:
```

gateways/python/fjagepy/Message.py
```diff
@@ -1,21 +1,75 @@
...
+    clazz_name = class_.__dict__.get('__clazz__')
+    if isinstance(clazz_name, str) and clazz_name:
+        return clazz_name
+    return "org.arl.fjage.Message"
+
+def _register_message_class(class_: Type["Message"], fqcn: Optional[str] = None) -> Type["Message"]:
```
dixslyf:
```suggestion
def _register_message_class(class_: type["Message"], fqcn: Optional[str] = None) -> type["Message"]:
```

gateways/python/fjagepy/Message.py
```diff
@@ -47,10 +112,14 @@ def __getattribute__(self, name):
         # Deal with attributes that are keywords in Python
         # by allowing them to be accessed with  a trailing underscore
         # e.g., msg.from --> msg.from_
-        if not name.startswith('_') and name.endswith('_'):
-            return object.__getattribute__(self, name[:-1])
+        normalized_name = _normalize_field_name(name)
+        if normalized_name != name:
+            return object.__getattribute__(self, normalized_name)
         return object.__getattribute__(self, name)
```
dixslyf: Since we're at it, I think `__getattr__` is a safer and more common way of doing this. Probably something like this:
```python
    def __getattr__(self, name):
        # Deal with attributes that are keywords in Python
        # by allowing them to be accessed with  a trailing underscore
        # e.g., msg.from --> msg.from_
        normalized_name = _normalize_field_name(name)
        if normalized_name != name:
            return getattr(self, normalized_name)
        raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
```
`getattr(self, normalized_name)` will raise an error automatically if the normalized name doesn't exist on the object.

dixslyf (changes requested):
Tests pass. Just some minor comments.

gateways/python/fjagepy/Message.py
notthetup: Good catch. I'll revert back.

dixslyf (approved):

PR #404, opened 2026-04-24
Fix import error when `numpy` is not installed in `fjagepy`

dixslyf:
`numpy` is not listed as a dependency of `fjagepy`. However, importing `fjagepy` causes an `ImportError` as it tries to import `numpy`. The problem is that, although we do check if `numpy` is installed, in the case where `numpy` is *not* installed, we are using `numpy.ndarray` in the type:
https://github.com/org-arl/fjage/blob/c09f1b76f9a89cba13baa0422060af2928faad40/gateways/python/fjagepy/Message.py#L33-L36

The only use of `_serialize_numpy_array` is here:
https://github.com/org-arl/fjage/blob/c09f1b76f9a89cba13baa0422060af2928faad40/gateways/python/fjagepy/Message.py#L136-L137

Since it already checks the `_NUMPY_AVAILABLE` flag, we can just omit the `_serialize_numpy_array()` definition in the case where `numpy` isn't installed.

### DIFF
--- a/gateways/python/fjagepy/Message.py
+++ b/gateways/python/fjagepy/Message.py
@@ -32,8 +32,6 @@ try:
 
 except ImportError:
     _NUMPY_AVAILABLE = False
-    def _serialize_numpy_array(value: numpy.ndarray, key: str, props: Dict):
-        return value
 
 def _normalize_field_name(name: str) -> str:
     """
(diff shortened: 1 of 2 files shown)


dixslyf:
@notthetup Alternatively, we can remove the `numpy` handling as we discussed over lunch.

notthetup (approved):

PR #414, opened 2026-06-25
Improve fjagejs JSDoc types

dixslyf:
Some of the JSDoc types for fjagejs were incorrect. There were also a few instances where the type was correct, but, unfortunately, TypeScript doesn't support all JSDoc tags: https://github.com/microsoft/TypeScript/issues/28730, https://github.com/microsoft/TypeScript/issues/20077.

This PR fixes incorrect JSDoc types and updates them so that they only use syntax that is recognised by TS.

As much as possible, I've tried to avoid making changes to actual code, but there a few instances where that's required due to type mismatches or to improve type narrowing.

This PR also adds a `typecheck` script to `package.json` so that running `pnpm run typecheck` will actually check that types are correct.

### DIFF
--- a/gateways/js/src/gateway.js
+++ b/gateways/js/src/gateway.js
@@ -19,7 +19,10 @@ const GATEWAY_DEFAULTS = {
   'returnNullOnFailedResponse': true
 };
 
+/** @type URL */
 let DEFAULT_URL;
+
+/** @type {{ fjage?: {gateways?: Gateway[]} } & Partial<typeof globalThis>} */
 let gObj = {};
 
 /**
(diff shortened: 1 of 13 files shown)


gateways/js/src/gateway.js
dixslyf, line 287: This is the one line where the type is incorrect. I am not sure what the best way to handle this is. The problem is that `JSONMessage`'s `agentID` attribute is defined to be an `AgentID`, but here we are setting it to a string.

If we change the `agentID` attribute to be `AgentID | string`, that fixes this part here, but causes problems in other code where we are calling `toJSON()` on `agentID` (`string` doesn't have a `toJSON()` method).

notthetup:
Also relevant is some work I did to align fjage.js to the fjage.py style of registering messages. But not yet complete.

https://github.com/org-arl/fjage/commits/fjagejs-registerMessageClass/

gateways/js/src/gateway.js
dixslyf: @notthetup Any suggestions on how to resolve this? Could we simply omit it or set it to `undefined`, and does `''` have a specific meaning in fjage?

PR #419, opened 2026-07-07
Support statically compiled FSM definitions

mchitre:
## Problem

The `FSMBuilder` Groovy DSL could not be used in `@CompileStatic` agent code — unqualified access to fields of the enclosing agent class from state closures (e.g. `count++`) failed static type checking, forcing FSM definitions to remain dynamically compiled.

## Root cause

The Groovy compiler adds synthetic `propertyMissing`/`methodMissing` bridge methods to **nested** Groovy classes (`FSMBuilder$StateBuilder`, `FSMBuilder$FSMState`). When user code compiles against the fjage **jar**, the static type checker reads these bridges from the bytecode and treats the `@DelegatesTo` delegate as resolving *any* unqualified property (as `Object`), shadowing the enclosing class's typed fields. This never shows up in same-project tests (different class-node resolution path), only downstream. Even `static` nested Groovy classes get the bridges, so the fix requires the delegate classes to not be nested Groovy classes.

## Fix

- Port `FSMBuilder` from Groovy to Java, preserving package, class, and nested-class names (source- and binary-compatible). Java nested classes carry no bridges; the existing `@DelegatesTo` annotations work identically.
- `after()` takes `Number` instead of `double`: the static checker does not coerce Groovy decimal literals (`BigDecimal`) to `double`, so `after(1.0)` now works both statically and dynamically.
- `FSMBehavior.State` gains `getAgent()`, `getCurrentState()`, `reenterState()`, and `trigger()` pass-throughs (matching the existing `setNextState`/`terminate`/`block` style) so these resolve on the closure delegate — `agent.add(...)` in a state closure previously did not type-check at all.
- Javadoc example uses `after(1)` (the `1.seconds` notation was never provided by fjage).

## Tests

- `testFSMBuilderStatic`: an `@CompileStatic` agent whose FSM exercises the full DSL — unqualified enclosing-class field access, captured locals, both `after` forms, `onEvent` with event info, `getCurrentState`, `trigger`, `agent` access, `terminate`. Since tests compile against the built classes, this is a true regression guard (it does not compile without the fix).
- `testFSMBuilder`: dynamic (non-static) DSL usage, guarding backward compatibility.
- `BasicTests` and `BehaviorTest` pass unchanged.

Verified as a drop-in replacement on an existing project by removing the dynamic compilation annotation from FSM definitions.


src/main/java/org/arl/fjage/groovy/FSMBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+  public static FSMBehavior build(@DelegatesTo(FSMBuilder.class) Closure<?> c) {
+    FSMBuilder fsm = new FSMBuilder();
+    c.setDelegate(fsm);
+    c.call();
+    return fsm;
+  }
+
+  public static class StateBuilder {
+
+    public FSMState state;
```
notthetup: Does it have to be public? Can it be at a lower access level?

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+ * <p>
+ * This class is implemented in Java (rather than Groovy) intentionally: the Groovy compiler
+ * adds synthetic {@code propertyMissing}/{@code methodMissing} bridge methods to nested
+ * Groovy classes, and the static type checker then resolves any unqualified property in a
+ * delegating closure against those bridges (as {@code Object}), breaking access to fields
+ * of the enclosing class in statically compiled FSM definitions.
+ */
+public class FSMBuilder extends FSMBehavior {
+
+  public static FSMBehavior build(@DelegatesTo(FSMBuilder.class) Closure<?> c) {
```
notthetup, line 75: Do we need a `strategy = Closure.DELEGATE_FIRST` here?

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    public void onEvent(Object name, @DelegatesTo(value=FSMState.class, strategy=Closure.DELEGATE_FIRST) Closure<?> c) {
+      c.setDelegate(state);
+      c.setResolveStrategy(Closure.DELEGATE_FIRST);
+      state.onEventClosures.put(name, c);
+    }
+
+    public WakerBehavior after(Number delay, @DelegatesTo(value=FSMState.class, strategy=Closure.DELEGATE_FIRST) final Closure<?> c) {
+      long millis = Math.round(delay.doubleValue()*1000);
```
notthetup, line 103: We should do a null check of `delay` and also ensure that it's positive.

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+      c.setDelegate(state);
+      c.setResolveStrategy(Closure.DELEGATE_FIRST);
+      WakerBehavior timer = new WakerBehavior(millis, new Runnable() {
```
notthetup: Use `new WakerBehavior(millis) {...}` for consistency?

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: And what should we do if it is null? Throw a null pointer exception? Which it will anyway?

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: Does Java support that??

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: Probably no need to be public. Will check.

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: The current one works fine. Any reason you think we should change?

mchitre: We shouldn't change it:

1. It's internally consistent, which is what static compilation depends on. `@DelegatesTo` defaults to `Closure.OWNER_FIRST`, and `build()` deliberately does not call `setResolveStrategy(...)`, so the runtime is also the default `OWNER_FIRST`. The invariant that makes `@CompileStatic` behave the same as dynamic dispatch is that the annotation's strategy matches the runtime resolve strategy — here both are `OWNER_FIRST`, so they agree. (Every inner method — `state()`, `onEnter`, etc. — sets `DELEGATE_FIRST` in both the annotation and via `setResolveStrategy`, so they're consistent too. The asymmetry is intentional, not an oversight.)
2. `OWNER_FIRST` is the correct choice for the top-level build block specifically. The build closure only calls `state(...)` on the delegate and declares locals like `long count = 0`. With `OWNER_FIRST`, enclosing-scope names (the agent's own locals/fields referenced at build-closure top level) resolve naturally, and state — which isn't an owner member — falls through to the `FSMBuilder` delegate. Switching to `DELEGATE_FIRST` would only change behavior in the edge case where the enclosing agent has a member literally named `state` or `build`, and in that case `DELEGATE_FIRST` would be the more surprising choice, not the safer one. The inner closures use `DELEGATE_FIRST` for a real reason (so `setNextState`/`block`/`terminate`/`getCurrentState`/`trigger`/`agent` on `FSMState` win) — that reason simply doesn't apply at the builder level.

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: Checked. Private will work, will update.

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: A zero or negative delay simply fires on the next action cycle — it doesn't crash, hang, or misbehave. And the null throws anyway, and we would want it thrown. I recommend no change.

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: Ah, I misunderstood what you meant. I thought you were asking `new WakerBehavior(millis, { ... })`. Yes, sure, we can update to use your recommendation.

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
notthetup: That's fine as long as that's the behaviour across Fjage. I remember we had some issues where negative delays caused exceptions in Fjage.

notthetup: Re: https://github.com/org-arl/unet/pull/1988

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: Fixed

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: Fixed

src/main/java/org/arl/fjage/groovy/FSMBuilder.java
mchitre: Tested that zero or negative fires immediately.

mchitre: OK, tracked down the issue from https://github.com/org-arl/unet/pull/1988. Relates to differences between real-time and discrete event platforms. Putting in a change to have uniform behavior across both and documenting it more clearly.

mchitre: Done, behavior is consistent now. Negative will always fire immediately.

notthetup (approved):

Issue #401, opened 2026-04-20
Revisiting timeouts in Remote/Slave Containers


notthetup:
`MasterContainer` has a default `TIMEOUT` of `15000` that used for all actions that performed, both for local Agents in the `MasterContainer` itself but ALSO for any remote Containers that are requesting the action remotely.

https://github.com/org-arl/fjage/blob/91ae1d7915be46c68c81c70b094d75ce89288d6f/src/main/java/org/arl/fjage/remote/MasterContainer.java#L294-L310

However, the Slave Container itself only uses a `TIMEOUT` of `2000` for any remote requests it makes.

https://github.com/org-arl/fjage/blob/91ae1d7915be46c68c81c70b094d75ce89288d6f/src/main/java/org/arl/fjage/remote/SlaveContainer.java#L215-L225

This would mean that it's possible for the `MasterContainer` to not respond within the timeout of the `SlaveContainer` depending on how slow the various remotes Container are.

Shouldn't this be the other way round? The default timeout for the `SlaveContainers` should be larger than what the `MasterContainer` uses to make sure that even when the `MasterContainer` times out on ONE of the remote Container, the querying SlaveContainer still get as much information as the `MasterContainer` has managed to collect so far.

---

Another issue is that the Gateway doesn't define what the default timeout for various Gateway Actions should be. This has caused Gateway implementations to diverge. We should define this so that we get similar default behaviour across the various Gateway implementations.

notthetup:
I propose updating the following.


## Containers

- On the `MasterContainer` default `TIMEOUT` for all directory queries (`SERVICES`, `AGENT_FOR_SERVICE`, `AGENTS_FOR_SERVICE`, `CONTAINS_AGENT`, `AGENTS`) to `5000`ms. This is large enough for most use cases. And with the new changes in https://github.com/org-arl/fjage/pull/398 we'll be doing this parallely so a query to each remote connection will get `5000`ms.
- We can leave the total timeout in `MasterContainer.initComplete` to `15000` since that has to wait for all the remotes to be connected and respond, and in some systems that might take time.
- On `SlaveContainer` we can change the `TIMEOUT` to `6000`ms. This is `1000`ms more than the query timeout for `MasterContainer`. That way even if the `MasterContainer` query times out, the `SlaveContainer` will get response. In certain cases like `AGENT_FOR_SERVICE` that could be enough, since as long as `MasterContainer` found a single Agent, the query can succeed even if one of the remote connections is taking too long.
[...]

PR #428, opened 2026-07-20
Update timeout defaults and add directory methods in Gateway spec

notthetup:
This PR updates the timeouts and adding missing directory methods in the Gatway spec. Re: #401

### DIFF
--- a/gateways/Gateways.md
+++ b/gateways/Gateways.md
@@ -52,7 +52,7 @@ All gateway agents should use names prefixed with `gateway-`.
 - Must not **block** if timeout is 0.
 - Must **block** indefinitely if timeout is -1.
 - Must **block** for timeout milliseconds otherwise.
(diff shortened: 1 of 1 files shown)


gateways/Gateways.md
```diff
@@ -52,15 +52,33 @@ All gateway agents should use names prefixed with `gateway-`.
 - Must not **block** if timeout is 0.
 - Must **block** indefinitely if timeout is -1.
 - Must **block** for timeout milliseconds otherwise.
-- Must default timeout to 1000 millisecond if not specified.
+- Must default timeout to 0 milliseconds if not specified.
```
mchitre, line 55: Why is this 0 ms when other versions of request() default to 1000 ms?

mchitre: Ah, receive() is 0 and request() is 1000. Makes sense. I misread.

mchitre (approved):

PR #431, opened 2026-07-23
Add support for statically checking/parsing scripts.

notthetup:
Add a mechanism to statically parse/check scripts using the ScriptEngine.

We add a new API `ScriptEngine.parse` which takes in a script and returns `MultipleCompilationErrorsException`.

The `ShellAgent` then accepts the new `ShellCheckReq` and responds with a `ShellCheckRsp` that contains a JSON-ified version of `MultipleCompilationErrorsException`.

### DIFF
--- a/src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
+++ b/src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
@@ -494,7 +494,6 @@ public class GroovyScriptEngine implements ScriptEngine {
 
   private RuntimeException rethrowParseFailure(Throwable ex) {
     if (ex instanceof GroovyBugError) ex = resolveGroovyBug(ex);
-    if (ex instanceof MultipleCompilationErrorsException) return (MultipleCompilationErrorsException) ex;
     if (ex instanceof RuntimeException) return (RuntimeException) ex;
     return new RuntimeException(ex);
   }
--- a/src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
+++ b/src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
@@ -137,16 +137,54 @@ public class GroovyScriptEngine implements ScriptEngine {
   public boolean isComplete(String cmd) {
     if (cmd == null || cmd.trim().length() == 0) return true;
     try {
-      groovy.parse(cmd);
-      return true;
+      return parse(cmd) == null;
     } catch (MultipleCompilationErrorsException ex) {
-      String s = ex.getMessage();
-      if (s == null) return true;
-      if (s.contains("unexpected token")) return false;
-      if (s.contains("expecting")) return false;
-      return true;
+      return !isIncomplete(ex);
     } catch (Throwable ex) {
       return true;
+    }
+  }
+
+  @Override
+  public MultipleCompilationErrorsException parse(String cmd) {
+    if (cmd == null || cmd.trim().length() == 0) return null;
+    synchronized(this) {
+      try {
+        busy = Thread.currentThread();
+        try {
+          groovy.parse(cmd);
+          return null;
+        } catch (MultipleCompilationErrorsException ex) {
+          return ex;
+        } catch (Throwable ex) {
+          throw rethrowParseFailure(ex);
+        }
+      } finally {
+        Thread.interrupted();
+        busy = null;
+      }
+    }
+  }
+
+  @Override
+  public MultipleCompilationErrorsException parse(File script) {
+    if (script == null) return null;
+    synchronized(this) {
+      try {
+        busy = Thread.currentThread();
+        try {
+          groovy.getClassLoader().clearCache();
+          groovy.parse(script);
+          return null;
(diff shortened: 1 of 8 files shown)


src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
```diff
@@ -136,17 +166,71 @@ public String getPrompt(boolean cont) {
...
+
+  @Override
+  public MultipleCompilationErrorsException parse(File script) {
+    if (script == null) return null;
+    synchronized(this) {
+      try {
+        busy = Thread.currentThread();
+        CHECK_BINDING.set(binding);
+        try {
+          groovyCheck.getClassLoader().clearCache();
```
mchitre: Needed? This will force a cache clear for everyone who is using the class loader.

src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
```diff
@@ -136,17 +166,71 @@ public String getPrompt(boolean cont) {
...
-      if (s == null) return true;
-      if (s.contains("unexpected token")) return false;
-      if (s.contains("expecting")) return false;
-      return true;
-    } catch (Throwable ex) {
-      return true;
+    // Use a dynamic (non-static) parse here: line-continuation detection only cares about
+    // syntactic completeness, and running the static check per keystroke would be wasteful and
+    // could misreport valid dynamic commands as errors.
+    synchronized(this) {
```
mchitre: Adding a lock here will block console while command is executing, potentially preventing user aborts.

src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
notthetup: Nope. Good catch.

src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
notthetup: Removed.

src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
```diff
@@ -135,21 +165,45 @@ public String getPrompt(boolean cont) {
...
+      return !isIncomplete(ex);
     } catch (Throwable ex) {
       return true;
     }
   }
 
+  @Override
+  public MultipleCompilationErrorsException parse(String cmd) {
+    if (cmd == null || cmd.trim().isEmpty()) return null;
+    CHECK_BINDING.set(binding);
```
mchitre, line 182: We might want a:
```
finally {
  CHECK_BINDING.remove();
}
```
block to release the binding since `CHECK_BINDING` is thread-local and doesn't garbage collect.

Same comment below for `parse(File script)`.

src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
```diff
@@ -115,6 +135,16 @@ private void init(GroovyClassLoader gcl) {
...
+    // Separate shell used only by parse() to statically check scripts without running them.
+    // It reuses the same (mutable) imports and script base class, and adds CompileStatic with a
+    // type-checking extension that tolerates shell binding variables and dynamic shell features.
+    CompilerConfiguration checkCompiler = new CompilerConfiguration();
+    checkCompiler.setScriptBaseClass(BaseGroovyScript.class.getName());
+    checkCompiler.addCompilationCustomizers(imports);
+    Map<String,Object> checkOpts = new HashMap<>();
+    checkOpts.put("extensions", new ConstantExpression("org/arl/fjage/shell/ShellStaticCheck.groovy"));
+    checkCompiler.addCompilationCustomizers(new ASTTransformationCustomizer(checkOpts, CompileStatic.class));
+    groovyCheck = new GroovyShell(gcl, binding, checkCompiler);
```
mchitre, line 147: We should perhaps use a different GroovyClassLoader for these checks and actually call `clearCache()` on it after every check. AFAIK, every `parse()` caches the compiled class, and so we may leak memory if we keep parsing. We don't actually need the compiled class, so calling `clearCache()` will discard it. But we can't do that on the `gcl` that is used for actually running the scripts, and so a separate check GroovyClassLoader might be needed so thay we can keep its cache empty.

src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
mchitre: My comment below re-explores this issue. I think the clearing may be a good idea, just not on the class loader used by everyone else.

src/main/groovy/org/arl/fjage/shell/GroovyScriptEngine.java
notthetup: OK. Let me see how we can do that.

notthetup: We'll need to clone the classloader that's being used for `exec` and then use that for `parse` and then throw it away after the check right?

PR #427, opened 2026-07-20
Run container queries in parallel

notthetup:
This pull request refactors the remote container communication and request handling system to improve concurrency, responsiveness, and robustness. The main changes include redesigning how requests and responses are tracked, introducing separate executor services for different types of tasks, and optimizing how the master container queries its slave containers. Additionally, timeouts and shutdown procedures have been improved for better reliability.

**Concurrency and Task Execution Improvements:**

- Introduced multiple executor services (`taskExecutor`, `directoryExecutor`, `sendExecutor`) in `ConnectionHandler` to separate general tasks, directory queries, and sending operations, improving parallelism and responsiveness. [[1]](diffhunk://#diff-9bdc333d13cef0928c209f8ac7ba4321db28f625076b86a776d31f0d9743df50L33-R44) [[2]](diffhunk://#diff-9bdc333d13cef0928c209f8ac7ba4321db28f625076b86a776d31f0d9743df50L150-R155) [[3]](diffhunk://#diff-9bdc333d13cef0928c209f8ac7ba4321db28f625076b86a776d31f0d9743df50R286-R293)

- Replaced the single `pool` executor with dedicated executors and updated shutdown logic to gracefully terminate all executors and notify pending requests on close. [[1]](diffhunk://#diff-9bdc333d13cef0928c209f8ac7ba4321db28f625076b86a776d31f0d9743df50L159-R165) [[2]](diffhunk://#diff-9bdc333d13cef0928c209f8ac7ba4321db28f625076b86a776d31f0d9743df50R286-R293) [[3]](diffhunk://#diff-9bdc333d13cef0928c209f8ac7ba4321db28f625076b86a776d31f0d9743df50R316-R339)

**Request/Response Handling Enhancements:**

[...]

PR #432, opened 2026-07-23
Classify gateway connections as lightweight, skip directory queries to them

mchitre:
Builds on #427 (base branch: `mastercontainer-parallel-queries`; retargets to master automatically when #427 merges).

## What

Master containers now identify gateway clients (Python/JS/C/Julia/Java `Gateway`) by the spec-mandated `gateway-` agent-name prefix (Gateways.md) and keep them lightweight:

- **Classification**: on a connection's first alive transition, the master sends one async `agents` probe. A response with exactly one `gateway-`-prefixed agent classifies the connection as a gateway (agent ID cached on the `ConnectionHandler`); any other response classifies it as a regular slave. Sticky for the connection lifetime; also fed opportunistically from normal `AGENTS` traffic if the probe misses.
- **Query suppression**: `querySlaves` skips classified gateways, so `agents`/`services`/`agentForService`/`agentsForService`/`containsAgent` are never sent to them.
- **Directory stays complete**: `getAgents()` appends the cached gateway agent IDs; `isDuplicate`/`canLocateAgent` answer gateway names locally with no network round-trip.
- **Name collision**: at classification, the gateway agent's name is checked against local agents and other gateways; on collision the master logs a warning, sends `action: shutdown` (stopping compliant clients' auto-reconnect), and closes the connection.

No wire-protocol change: the `gateway-` prefix is already normative in `gateways/Gateways.md`; the spec is updated to reserve the prefix explicitly, document the lightweight-connection consequence, forbid gateways from registering services, and specify shutdown-on-collision.

## Docs note

`docs/gateways.qmd` only exists on master (Quarto migration postdates this branch point), so the user-facing paragraph is in `src/sphinx/remote.rst` here; it should move to `docs/gateways.qmd` when this stack rebases onto master.

## Tests

New `GatewayConnectionTest` (7 cases) using a raw-TCP fake gateway for deterministic names: classification, query suppression, cached directory visibility + fast local lookup, collision with local agent, collision between gateways, empty-at-probe slave not misclassified (late agents still visible), and end-to-end with a real Java `Gateway`. Verified green alongside #427's `MasterContainerQueryTest`, `BasicTests`, and `fjagejsTest` (real JS gateway).


notthetup (approved):

mchitre:
Will wait for #427 to merge before merging this.

PR #435, opened 2026-07-30
Add SerialConnector for serial port communication to fjagepy

notthetup:
This pull request adds support for connecting to fjåge containers over serial ports in the Python gateway, alongside the existing TCP support. It introduces a new `SerialConnector` class that allows connecting to fjåge Master container from a fjage.py Gateway.

This PR also changes the API for `Gateway()` where it now takes in an optional `Connector` instance (instead of type in the past). This simplifies the logic inside the Gateway. But since this breaks the API, this PR also uprevs the MAJOR version for fjage.py.

### DIFF
--- a/gateways/python/fjagepy/SerialConnector.py
+++ b/gateways/python/fjagepy/SerialConnector.py
@@ -15,14 +15,14 @@ class SerialConnector(Connector):
     Requires the optional `pyserial` dependency (`pip install fjagepy[serial]`).
     """
 
-    def __init__(self, devname: str, baud: int = 9600, reconnect_delay: float = -2):
+    def __init__(self, devname: str, reconnect_delay: float, baud: int = 9600):
         self.devname = devname
         if not self.devname or not isinstance(self.devname, str) or self.devname.strip() == "":
             raise ValueError("devname must be a non-empty string")
         self.baud = baud
         if not isinstance(self.baud, int) or self.baud <= 0:
             raise ValueError("baud must be a positive integer")
-        self.reconnect_delay = kwargs.get("reconnect_delay", None)
+        self.reconnect_delay = reconnect_delay
         if not isinstance(self.reconnect_delay, (int, float)) or self.reconnect_delay < -1:
             raise ValueError("reconnect_delay must be a non-negative number or -1 for no reconnect")
 
--- a/gateways/python/fjagepy/SerialConnector.py
+++ b/gateways/python/fjagepy/SerialConnector.py
@@ -1,4 +1,3 @@
-import socket
 import logging
 import threading
 import time
@@ -10,23 +9,31 @@ logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
 
+class SerialConnector(Connector):
+    """Simple serial port connector using pyserial.
 
-class TCPConnector(Connector):
-    """Simple TCP connector using synchronous sockets."""
+    Requires the optional `pyserial` dependency (`pip install fjagepy[serial]`).
+    """
 
-    def __init__(self, **kwargs: Any ):
-        self.host = kwargs.get("host")
-        if not self.host or not isinstance(self.host, str) or self.host.strip() == "":
-            raise ValueError("host must be a non-empty string")
-        self.port:int = kwargs.get("port", 0)
-        if not isinstance(self.port, int) or not (0 < self.port < 65536):
-            raise ValueError("port must be an integer 1-65535")
-        self.reconnect_delay:int = kwargs.get("reconnect_delay", -2)
+    def __init__(self, **kwargs: Any):
+        self.devname = kwargs.get("devname")
+        if not self.devname or not isinstance(self.devname, str) or self.devname.strip() == "":
+            raise ValueError("devname must be a non-empty string")
+        self.baud: int = kwargs.get("baud", 9600)
+        if not isinstance(self.baud, int) or self.baud <= 0:
+            raise ValueError("baud must be a positive integer")
+        self.reconnect_delay = kwargs.get("reconnect_delay", -2)
         if not isinstance(self.reconnect_delay, (int, float)) or self.reconnect_delay < -1:
             raise ValueError("reconnect_delay must be a non-negative number or -1 for no reconnect")
 
-        # Socket and connection state
-        self._socket: Optional[socket.socket] = None
+        try:
+            import serial  # type: ignore[import-untyped]  # local import, so that pyserial stays an optional dependency
+        except ImportError as e:
+            raise ImportError("SerialConnector requires pyserial: pip install pyserial") from e
+        self._serial = serial
+
+        # Port and connection state
+        self._port: Optional[Any] = None
         self._connected = False
         self._callback: Optional[Callable[[List[str]], None]] = None
 
@@ -42,17 +49,16 @@ class TCPConnector(Connector):
         self._callback = callback
 
     def connect(self) -> None:
-        """Establish the connection."""
+        """Open the serial port."""
         with self._lock:
             if self._connected:
                 return
 
             try:
-                self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-                self._socket.settimeout(10.0)  # 10 second connection timeout
-                self._socket.connect((self.host, self.port))
+                # serial_for_url() accepts device names as well as URLs (eg. loop://)
+                self._port = self._serial.serial_for_url(self.devname, baudrate=self.baud, timeout=0.1)
                 self._connected = True
-                logger.debug(f"Connected to {self.host}:{self.port}")
+                logger.debug(f"Connected to {self.devname}@{self.baud}")
 
                 # Start read thread if callback is set
                 if self._callback:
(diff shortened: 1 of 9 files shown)


gateways/python/fjagepy/SerialConnector.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+
+    def __init__(self, **kwargs: Any):
+        self.devname = kwargs.get("devname")
+        if not self.devname or not isinstance(self.devname, str) or self.devname.strip() == "":
+            raise ValueError("devname must be a non-empty string")
+        self.baud: int = kwargs.get("baud", 9600)
+        if not isinstance(self.baud, int) or self.baud <= 0:
+            raise ValueError("baud must be a positive integer")
+        self.reconnect_delay = kwargs.get("reconnect_delay", -2)
```
dixslyf: Is setting to `-2` as the default intentional? If the intention is to force the caller to pass in `reconnect_delay`, then I suggest defaulting to `None` instead.

gateways/python/fjagepy/SerialConnector.py
```diff
@@ -0,0 +1,177 @@
...
+        if not isinstance(self.reconnect_delay, (int, float)) or self.reconnect_delay < -1:
+            raise ValueError("reconnect_delay must be a non-negative number or -1 for no reconnect")
```
dixslyf: Not much reason to use `**kwargs` since we're not doing anything dynamic with the kwargs. Explicitly declaring the arguments will have better type inference / autocomplete:
```suggestion
    def __init__(self, devname: str, baud: int = 9600, reconnect_delay: float = -2):
        self.devname = devname
        if not self.devname or not isinstance(self.devname, str) or self.devname.strip() == "":
            raise ValueError("devname must be a non-empty string")
        self.baud = baud
        if not isinstance(self.baud, int) or self.baud <= 0:
            raise ValueError("baud must be a positive integer")
        self.reconnect_delay = reconnect_delay
        if not isinstance(self.reconnect_delay, (int, float)) or self.reconnect_delay < -1:
            raise ValueError("reconnect_delay must be a non-negative number or -1 for no reconnect")
```

dixslyf: We should do the same for `TCPConnector`, but I'll raise another PR for that.

gateways/python/fjagepy/SerialConnector.py
```diff
@@ -0,0 +1,177 @@
...
+
+                # Start read thread if callback is set
+                if self._callback:
+                    self._stop_event.clear()
+                    self._read_thread = threading.Thread(target=self._read_loop, daemon=True)
+                    self._read_thread.start()
+
+            except Exception as e:
+                self._cleanup_port()
+                raise ConnectionError(f"Failed to connect to {self.devname}@{self.baud}: {e}")
```
dixslyf, line 85:
```suggestion
            except Exception as e:
                self._connected = False
                self._cleanup_port()
                raise ConnectionError(f"Failed to connect to {self.devname}@{self.baud}: {e}")
```

gateways/python/fjagepy/SerialConnector.py
```diff
@@ -0,0 +1,177 @@
...
+        if self._port:
+            try:
+                self._port.close()
+            except Exception:
+                pass
+            self._port = None
+
+    def _read_loop(self) -> None:
+        """Background thread that reads data and calls the callback."""
+        buffer = ""
```
dixslyf:
```suggestion
        buffer: str = ""
```
Seems like pyrefly infers the type as `Literal['']`, which breaks some of the type inference later in the loop.

gateways/python/fjagepy/SerialConnector.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def _attempt_reconnect(self) -> None:
+        """Attempt to reopen the port with delay."""
+        while not self._stop_event.is_set() and self.reconnect_delay >= 0:
+            try:
+                logger.debug(f"Attempting to reconnect in {self.reconnect_delay}s...")
+                time.sleep(self.reconnect_delay)
+
+                if self._stop_event.is_set():
+                    break
```
dixslyf:
```suggestion
                if self._stop_event.wait(timeout=self.reconnect_delay):
                    break 
```

Instead of making the thread sleep and only seeing the set event `reconnect_delay` seconds later (which could be well after the `self._read_thread.join(timeout=2.0)` timeout), better to use `.wait()` on the event.

dixslyf (changes requested):

gateways/python/fjagepy/SerialConnector.py
dixslyf: Actually, since the args are just normal arguments now, you can just not set a default value.

gateways/python/fjagepy/SerialConnector.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+
+    def __init__(self, devname: str, baud: int = 9600, reconnect_delay: float = -2):
+        self.devname = devname
+        if not self.devname or not isinstance(self.devname, str) or self.devname.strip() == "":
+            raise ValueError("devname must be a non-empty string")
+        self.baud = baud
+        if not isinstance(self.baud, int) or self.baud <= 0:
+            raise ValueError("baud must be a positive integer")
+        self.reconnect_delay = kwargs.get("reconnect_delay", None)
```
dixslyf: `kwargs` is not defined anywhere, so this will fail. Let's just not set a default value for `reconnect_delay` to force the caller to specify a value.


```suggestion
    def __init__(self, devname: str, reconnect_delay: float, baud: int = 9600):
        self.devname = devname
        if not self.devname or not isinstance(self.devname, str) or self.devname.strip() == "":
            raise ValueError("devname must be a non-empty string")
        self.baud = baud
        if not isinstance(self.baud, int) or self.baud <= 0:
            raise ValueError("baud must be a positive integer")
        self.reconnect_delay = reconnect_delay
```

dixslyf (changes requested):

dixslyf:
I pushed a commit to resolve the `kwargs` issue as there was something urgent that needed this to work.

gateways/python/fjagepy/SerialConnector.py
dixslyf: I pushed the above change as there was something urgent that needed this to work.

gateways/python/fjagepy/SerialConnector.py
```diff
@@ -0,0 +1,176 @@
...
+logger.addHandler(logging.NullHandler())
+
+
+class SerialConnector(Connector):
+    """Simple serial port connector using pyserial.
+
+    Requires the optional `pyserial` dependency (`pip install fjagepy[serial]`).
+    """
+
+    def __init__(self, devname: str, reconnect_delay: float, baud: int = 9600):
```
dixslyf: As discussed offline, let's set a default value for `reconnect_delay` after all.

dixslyf:
```suggestion
    def __init__(self, devname: str, reconnect_delay: float, baud: int = 9600, **kwargs: Any):
```

Adding `**kwargs` at the end should allow type checkers to still type check properly while allowing arbitrary arguments that are ignored.

dixslyf: If we're passing an instance of a Connector to the Gateway, then we don't need this anymore.

gateways/python/fjagepy/TCPConnector.py
```diff
@@ -14,16 +14,16 @@
...
-    def __init__(self, **kwargs: Any ):
-        self.host = kwargs.get("host")
-        if not self.host or not isinstance(self.host, str) or self.host.strip() == "":
-            raise ValueError("host must be a non-empty string")
-        self.port:int = kwargs.get("port", 0)
-        if not isinstance(self.port, int) or not (0 < self.port < 65536):
-            raise ValueError("port must be an integer 1-65535")
-        self.reconnect_delay:int = kwargs.get("reconnect_delay", -2)
-        if not isinstance(self.reconnect_delay, (int, float)) or self.reconnect_delay < -1:
-            raise ValueError("reconnect_delay must be a non-negative number or -1 for no reconnect")
```
dixslyf, line 26: We should still validate the constructor arguments.

gateways/python/fjagepy/SerialConnector.py
```diff
@@ -0,0 +1,176 @@
...
+    def __init__(self, devname: str, baud: int = 9600, reconnect_delay: float = 5) -> None:
+        """
+        Args:
+            devname : serial device name (e.g. '/dev/ttyUSB0'), or a pyserial URL (e.g. 'loop://').
+            baud : baud rate. Defaults to 9600.
+            reconnect_delay : seconds to wait before reconnecting, or -1 to disable reconnection. Defaults to 5.
+        """
+        self.devname = devname
+        self.baud = baud
+        self.reconnect_delay = reconnect_delay
```
dixslyf, line 40: We should re-add validation of the constructor arguments.

dixslyf (changes requested):

gateways/python/fjagepy/SerialConnector.py
```diff
@@ -0,0 +1,189 @@
...
+    def __init__(self, devname: str, baud: int = 9600, reconnect_delay: float = 5) -> None:
+        """
+        Args:
+            devname : serial device name (e.g. '/dev/ttyUSB0'), or a pyserial URL (e.g. 'loop://').
+            baud : baud rate. Defaults to 9600.
+            reconnect_delay : seconds to wait before reconnecting, or -1 to disable reconnection. Defaults to 5.
+        """
+        if not isinstance(devname, str) or not devname.strip():
+            raise ValueError("devname must be a non-empty string")
+        if isinstance(baud, bool) or not isinstance(baud, int) or baud <= 0:
```
dixslyf, line 28: `bool` check is redundant. Same for other parameters in `SerialConnector` and `TCPConnector`.

dixslyf (approved):

gateways/python/fjagepy/SerialConnector.py
notthetup: Apparently that's a Python thing: 

> Because Python makes bool a subclass of int:
> `isinstance(True, int)  # True`
> Without the explicit isinstance(baud, bool) check, baud=True would pass isinstance(baud, int) and baud > 0, then be used as baud rate 1. The bool check ensures only actual positive integer baud rates are accepted.

dixslyf: Wow! TIL. Let's leave it then.

Issue #437, opened 2026-08-01
Gateway.receive from non-gateway thread can hang forever: untimed wait, wakeup lost if behavior dies


mchitre:
Affects fjage 2.6.0 (and the same code is present on master).

## Problem

`Gateway.receive(MessageFilter, long)` called from a non-gateway thread parks the caller in an **untimed** wait, with the wakeup delivered only if the internal behavior completes normally (`Gateway.init()`):

```java
synchronized (sync) {
  rsp = null;
  try {
    add(new OneShotBehavior() {
      @Override
      public void action() {
        rsp = receive(filter, timeout);          // timed receive on the gateway agent thread
        synchronized (sync) { sync.notify(); }   // wakeup ONLY on normal completion
      }
    });
    sync.wait();                                 // <-- untimed
  } catch (InterruptedException ex) { ... }
  return rsp;
}
```

The timeout requested by the caller is honored only inside the behavior. If the behavior terminates abnormally — an unchecked exception inside `action()` (e.g. #438 thrown from the receive plumbing on connection loss), or the behavior being discarded because the agent/container is shutting down — `sync.notify()` is never reached and the caller blocks **forever**, in violation of the timeout contract of `receive()`/`request()`.

## Observed

Reproduced intermittently while driving a UnetStack RealTimePlatform simulation through `UnetSocket` (which delegates to `Gateway.receive` with a 60 s timeout): the caller thread was found parked in `Object.wait()` (state WAITING, i.e. untimed) at `Gateway$1.receive`, while the gateway agent thread was idle in `Agent.block()` — the behavior had come and gone without notifying. Only interrupting the waiting thread (`UnetSocket.cancel()`) unblocks it.

## Suggested fix

Two independent hardenings, ideally both:

1. wrap the behavior action in `try/finally` so `sync.notify()` fires on all exit paths;
2. make the caller-side wait timed as a backstop: loop on `sync.wait(remaining)` bounded by the requested timeout (plus slack), returning null on expiry.

With (1) alone, a discarded (never-run) behavior still hangs the caller; (2) covers that case too.

PR #439, opened 2026-08-01
Avoid NPE when master connection is lost

### LINKED ISSUE #438: NPE in SlaveContainer.checkAuthFailure when connection to master is lost
Affects fjage 2.6.0 (and the same code is present on master).

## Problem

`SlaveContainer.checkAuthFailure()` dereferences `master` without a guard:

```java
public void checkAuthFailure(String id) {
  if (master.checkAuthFailure(id)) throw new AuthFailureException();
}
```

When the TCP connection to the master container is lost, `master` is no longer valid and this throws a raw `NullPointerException`. It is called from `Gateway.receive(Message, long)` (both before and after the wait), so API users of `Gateway.receive`/`request` get an NPE out of the public API instead of a clean `null` return or a meaningful exception.

## Observed

```
INFO: Connection to localhost:1102 lost
Caught: java.lang.NullPointerException
	at org.arl.fjage.remote.SlaveContainer.checkAuthFailure(SlaveContainer.java:135)
	at org.arl.fjage.remote.Gateway.receive(Gateway.java:249)
	at org.arl.fjage.remote.Gateway.request(Gateway.java:273)
```

(fjage 2.6.0, gateway connected to a UnetStack RealTimePlatform simulation; connection dropped mid-`request`.)

## Impact

Beyond the API leak, an unchecked exception thrown at an unexpected point in the receive plumbing is one of the triggers for #437 (permanent hang of a concurrent `Gateway.receive`) when it fires on the gateway agent thread.

## Suggested fix

Guard for a null/closed `master` and treat it as "no auth failure" (letting the receive return null and the connection-loss surface through the existing reconnect/close handling), e.g.:

```java
public void checkAuthFailure(String id) {
  ConnectionHandler m = master;
  if (m != null && m.checkAuthFailure(id)) throw new AuthFailureException();
}
```

mchitre:
## Summary

Fixes #438.

`SlaveContainer.checkAuthFailure()` dereferenced the volatile `master` field without a null guard. `master` is nulled by the connection-manager thread when the TCP connection to the master container is lost, so a raw `NullPointerException` escaped through the public `Gateway.receive()`/`request()` API exactly when a connection dropped mid-call:

```
INFO: Connection to localhost:1102 lost
java.lang.NullPointerException
  at org.arl.fjage.remote.SlaveContainer.checkAuthFailure
  at org.arl.fjage.remote.Gateway.receive
  at org.arl.fjage.remote.Gateway.request
```

This NPE (when thrown on the gateway agent thread) is also one of the triggers for the untimed-wait hang in #437, which will be fixed separately.

## Changes

- `checkAuthFailure()`: snapshot `master` into a local and treat null as "no auth failure" — connection loss then surfaces as a null response through the existing reconnect/close handling, which is what callers already expect.
- Audit of the rest of `SlaveContainer` for the same pattern: `authenticate()`, `isDuplicate()`, `initComplete()` and `shutdown()` all had check-then-reread races on the volatile field; they now snapshot into a local, matching the pattern already used in `send()` and the directory-query methods.

## Tests

New `SlaveContainerConnectionLossTest` uses `ThrottlingTcpProxy` to drop the slave→master TCP connection abruptly (RST, no SIGN_OFF — a graceful `MasterContainer.shutdown()` sends a SHUTDOWN action and doesn't reproduce the bug):

- `requestAfterConnectionLossReturnsNull` — drop the connection, then `checkAuthFailure()` must not throw and `request()` must return null.
- `requestDuringConnectionLossReturnsNull` — drop the connection while a `request()` is waiting; it must return null, never NPE.

Both tests fail with the exact field-observed NPE before the fix and pass after. Also ran `org.arl.fjage.remote.*`, `BasicTests` and `FirewallTests` — all pass.

### DIFF
--- a/src/main/java/org/arl/fjage/remote/SlaveContainer.java
+++ b/src/main/java/org/arl/fjage/remote/SlaveContainer.java
@@ -119,10 +119,11 @@ public class SlaveContainer extends RemoteContainer {
    * @return true if authenticated, false otherwise.
    */
   public boolean authenticate(String creds) {
-    if (master == null) return false;
+    ConnectionHandler localMaster = master;
+    if (localMaster == null) return false;
     JsonMessage rq = JsonMessage.createActionRequest(Action.AUTH);
     rq.creds = creds;
-    JsonMessage rsp = master.request(rq, REQUEST_TIMEOUT);
+    JsonMessage rsp = localMaster.request(rq, REQUEST_TIMEOUT);
     return rsp != null && rsp.auth != null && rsp.auth;
   }
 
@@ -132,7 +133,8 @@ public class SlaveContainer extends RemoteContainer {
    * INTERNAL USE ONLY.
    */
   public void checkAuthFailure(String id) {
-    if (master.checkAuthFailure(id)) throw new AuthFailureException();
+    ConnectionHandler localMaster = master;    // master may be nulled on connection loss
+    if (localMaster != null && localMaster.checkAuthFailure(id)) throw new AuthFailureException();
   }
 
   /////////////// Container interface methods to override
@@ -140,10 +142,11 @@ public class SlaveContainer extends RemoteContainer {
   @Override
   protected boolean isDuplicate(AgentID aid) {
     if (super.isDuplicate(aid)) return true;
-    if (master == null) return false;
+    ConnectionHandler localMaster = master;
+    if (localMaster == null) return false;
     JsonMessage rq = JsonMessage.createActionRequest(Action.CONTAINS_AGENT);
     rq.agentID = aid;
-    JsonMessage rsp = master.request(rq, DIRECTORY_QUERY_TIMEOUT);
+    JsonMessage rsp = localMaster.request(rq, DIRECTORY_QUERY_TIMEOUT);
     return rsp != null && rsp.answer != null && rsp.answer;
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/arl/fjage/remote/SlaveContainer.java
```diff
@@ -132,18 +133,20 @@ public boolean authenticate(String creds) {
    * INTERNAL USE ONLY.
    */
   public void checkAuthFailure(String id) {
-    if (master.checkAuthFailure(id)) throw new AuthFailureException();
+    ConnectionHandler localMaster = master;    // master may be nulled on connection loss
```
notthetup, line 136: Is this really necessary here? The only thing we're doing is `localMaster.checkAuthFailure(id)`

src/main/java/org/arl/fjage/remote/SlaveContainer.java
```diff
@@ -261,7 +266,8 @@ protected void initComplete() {
   @Override
   public void shutdown() {
     quit = true;
-    if (master != null) master.close();
+    ConnectionHandler localMaster = master;
```
notthetup, line 269: Is this really necessary here? The only thing we're doing is `localMaster.close()`

src/main/java/org/arl/fjage/remote/SlaveContainer.java
mchitre: You mean copying to local? Yes, because otherwise the global can change between the check and actual call. In fact, that was the bug that this PR grew out of.

src/main/java/org/arl/fjage/remote/SlaveContainer.java
mchitre: Same as above.

notthetup (approved):

PR #442, opened 2026-08-04
Message observer web-ui for debugging fjåge applications

mchitre:
## What

fjåge offers no way to watch agents talk to each other. Logs show what each agent chose to log, and a shell sees only the messages addressed to it. When you are debugging a protocol between three agents, you want the wire, and there is no wire to look at.

This adds `org.arl.fjage.observer` — a self-contained package. Added to a container, `Observer` observes every message sent in that container and serves a web interface that draws the traffic as a sequence diagram.

```groovy
container.add 'observer', new org.arl.fjage.observer.Observer()
```

The web interface is served by the application's own web server if it has one, and on port 8081 otherwise. The URL is logged at startup and readable as the agent's `url` parameter.

image

## Design notes

- **Observation is transparent.** The listener never consumes a message, and its whole body is wrapped in a `catch (Throwable)` — `Container.send()` does not guard listener calls, so an exception there would break delivery for the container.
- **The observer reports messages _sent_.** The listener runs before delivery is attempted, so it cannot say whether a message was delivered. A message published to a topic appears once, addressed to the topic, not once per subscriber.
- **One container per observer.** Traffic between two agents inside a slave never reaches the master. The container name is shown in the UI to head off the "why can't I see agent X" question.
- **Serialization is inline, on the sending thread.** Delivery within a container is by reference, so a queued `Message` could be mutated by its recipient before being serialized. Messages go out one JSON object per line, in the same envelope the gateways speak (`JsonMessage` + `Action.SEND`), with `time`, `ptime` and `seq` spliced onto the front — so any user-defined `Message` subclass round-trips via the existing adapters.
- **Filtering is split by intent.** Choosing which endpoints to *show* is local to the browser and retrospective — it reveals messages already buffered there. *Dropping* an endpoint, or excluding message classes by regex, is applied in the container before serialization, so those messages are never sent to any browser and undoing the rule does not bring them back. Container-side state is echoed to every client so open UIs agree, and a newly connected client adopts the shared filter rather than pushing its own — only an explicit apply or drop mutates it.
- **Endpoint discovery runs before filtering**, so the endpoint list stays complete even while traffic is being dropped, and an excluded endpoint can always be un-excluded.
- **Simulation-safe.** Both wall clock and platform time are on every event, and periodic emission is guarded by a wall-clock check, since virtual time can race ahead under `DiscreteEventSimulator`.

## UI

Plain HTML/CSS/JS served from the jar — no build step, no bundler, no CDN.

Per-endpoint message counts, a time gutter with inter-message deltas, reply correlation via `inReplyTo` (click a message to highlight what it replies to and any replies to it), drag to reorder lifelines, ↑/↓ keyboard navigation, pause/resume with a proper display freeze, a bounded in-browser buffer, and light/dark themes with selections and layout persisted in `localStorage`.

## Scope

Deliberately a first cut: no subscription/topology view, no recording and replay, no message table, no multi-source merge. Nothing outside the new package is touched.

### DIFF
--- a/src/main/java/org/arl/fjage/observer/ObserverAgent.java
+++ b/src/main/java/org/arl/fjage/observer/ObserverAgent.java
@@ -119,7 +119,7 @@ public class ObserverAgent extends Agent implements MessageListener {
   protected final String context;
   protected final Gson gson = new Gson();
   protected final Object wlock = new Object();
-  protected final Map<String,Endpoint> endpoints = new ConcurrentHashMap<String,Endpoint>();
+  protected final Map<String,Endpoint> endpoints = new ConcurrentHashMap<>();
 
   protected WebServer server = null;
   protected WebSocketHubConnector conn = null;
@@ -186,13 +186,10 @@ public class ObserverAgent extends Agent implements MessageListener {
       log.warning("Observer web interface resources not found at "+RESOURCE);
     conn = new WebSocketHubConnector(port, context+"/ws");
     out = conn.getOutputStream();
-    conn.setConnectionListener(new ConnectionListener() {
-      @Override
-      public void connected(Connector connector) {
-        // bring a newly connected client up to date
-        publishState();
-        publishEndpoints(null);
-      }
+    conn.setConnectionListener(connector -> {
+      // bring a newly connected client up to date
+      publishState();
+      publishEndpoints(null);
     });
     startControlThread();
     if (c != null) c.addListener(this);
@@ -272,7 +269,7 @@ public class ObserverAgent extends Agent implements MessageListener {
     } catch (Throwable ex) {
       // an exception here would break message delivery for the whole
       // container, since Container.send() does not guard listener calls
-      log.log(Level.WARNING, "Observer error: "+ex.toString(), ex);
+      log.log(Level.WARNING, "Observer error: "+ ex, ex);
     }
     return false;
   }
@@ -310,7 +307,7 @@ public class ObserverAgent extends Agent implements MessageListener {
         Endpoint prev = endpoints.putIfAbsent(name, ep = new Endpoint(aid.isTopic()));
         if (prev != null) ep = prev;
         else {
-          if (fresh == null) fresh = new ArrayList<String>(2);
+          if (fresh == null) fresh = new ArrayList<>(2)
           fresh.add(name);
         }
       }
@@ -367,7 +364,7 @@ public class ObserverAgent extends Agent implements MessageListener {
         publish(json);
       }
     } catch (Throwable ex) {
-      log.log(Level.WARNING, "Observer error: "+ex.toString(), ex);
+      log.log(Level.WARNING, "Observer error: "+ ex, ex);
     }
   }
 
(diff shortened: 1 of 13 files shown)


src/main/java/org/arl/fjage/observer/ObserverAgent.java
```diff
@@ -0,0 +1,592 @@
...
+  public static final int DEFAULT_MAX_RATE = 500;
+
+  protected static final String RESOURCE = "/org/arl/fjage/observer/web";
+  protected static final long STATS_INTERVAL = 2000;
+
+  protected int port;
+  protected final String context;
+  protected final Gson gson = new Gson();
+  protected final Object wlock = new Object();
+  protected final Map<String,Endpoint> endpoints = new ConcurrentHashMap<String,Endpoint>();
```
notthetup:
```suggestion
  protected final Map<String,Endpoint> endpoints = new ConcurrentHashMap<>();
```

src/main/java/org/arl/fjage/observer/ObserverAgent.java
```diff
@@ -0,0 +1,592 @@
...
+    conn = new WebSocketHubConnector(port, context+"/ws");
+    out = conn.getOutputStream();
+    conn.setConnectionListener(new ConnectionListener() {
+      @Override
+      public void connected(Connector connector) {
+        // bring a newly connected client up to date
+        publishState();
+        publishEndpoints(null);
+      }
+    });
```
notthetup:
```suggestion
    conn.setConnectionListener(connector -> {
      // bring a newly connected client up to date
      publishState();
      publishEndpoints(null);
    });
```

src/main/java/org/arl/fjage/observer/ObserverAgent.java
```diff
@@ -0,0 +1,592 @@
...
+          dropped++;
+          return false;
+        }
+        count++;
+        publish(serialize(msg, t, ++seq));
+      }
+    } catch (Throwable ex) {
+      // an exception here would break message delivery for the whole
+      // container, since Container.send() does not guard listener calls
+      log.log(Level.WARNING, "Observer error: "+ex.toString(), ex);
```
notthetup:
```suggestion
      log.log(Level.WARNING, "Observer error: "+ ex, ex);
```

src/main/java/org/arl/fjage/observer/ObserverAgent.java
```diff
@@ -0,0 +1,592 @@
...
+    for (AgentID aid: aids) {
+      String name = endpoint(aid);
+      if (name == null || name.equals(last)) continue;    // count a self-message once
+      last = name;
+      Endpoint ep = endpoints.get(name);
+      if (ep == null) {
+        Endpoint prev = endpoints.putIfAbsent(name, ep = new Endpoint(aid.isTopic()));
+        if (prev != null) ep = prev;
+        else {
+          if (fresh == null) fresh = new ArrayList<String>(2);
```
notthetup:
```suggestion
          if (fresh == null) fresh = new ArrayList<>(2)
```

src/main/java/org/arl/fjage/observer/ObserverAgent.java
```diff
@@ -0,0 +1,592 @@
...
+  }
+
+  /** Publishes a line to all connected web clients, swallowing any error. */
+  protected void publishSafely(String json) {
+    try {
+      synchronized (wlock) {
+        publish(json);
+      }
+    } catch (Throwable ex) {
+      log.log(Level.WARNING, "Observer error: "+ex.toString(), ex);
```
notthetup:
```suggestion
      log.log(Level.WARNING, "Observer error: "+ ex, ex);
```

src/main/java/org/arl/fjage/observer/ObserverAgent.java
```diff
@@ -0,0 +1,592 @@
...
+              handleControl(s);
+            } catch (Throwable ex) {
+              log.log(Level.WARNING, "Bad observer command: "+s, ex);
+            }
+          }
+        } catch (Throwable ex) {
+          if (conn != null) log.log(Level.WARNING, "Observer control thread died: "+ex.toString(), ex);
+        }
+      }
+    }, getClass().getSimpleName()+":ctrl");
```
notthetup:
```suggestion
    ctrlThread = new Thread(() -> {
      try {
        BufferedReader in = new BufferedReader(new InputStreamReader(conn.getInputStream(), StandardCharsets.UTF_8));
        String s;
        while ((s = in.readLine()) != null) {
          s = s.trim();
          if (s.isEmpty()) continue;
          try {
            handleControl(s);
          } catch (Throwable ex) {
            log.log(Level.WARNING, "Bad observer command: "+s, ex);
          }
        }
      } catch (Throwable ex) {
        if (conn != null) log.log(Level.WARNING, "Observer control thread died: "+ ex, ex);
      }
    }, getClass().getSimpleName()+":ctrl");
```

notthetup:
1. Opening another observer tab can silently replace the global container filter.

On every WebSocket connection, the UI immediately sends its locally persisted filter before requesting current state (observer.js). The server applies it globally (ObserverAgent.java). A stale or empty browser profile therefore clears another user’s active exclusions, or reintroduces exclusions after an observer restart. New clients should request/apply state first; only explicit Apply/drop actions should mutate the shared filter.

2. Statistics may be inaccurate under concurrent sends.

`count` and `dropped` are plain `long`s written from arbitrary sender threads and read from ticker/WebSocket threads (ObserverAgent.java:137, ObserverAgent.java:412). Reads can be stale and, on Java 8, non-atomic. Use `AtomicLong` or synchronize statistic snapshots.

mchitre:
Both addressed in 99dec458.

1. A newly connected client no longer sends its filter on connect — it just requests state and adopts the shared filter. Only an explicit Apply (or a drop/undrop click) mutates the container-side filter. Drop rules are also no longer persisted in `localStorage`, since the container is authoritative and a persisted copy could only ever be stale.

2. `count` and `dropped` are now `AtomicLong`s, snapshotted once per stats tick so the change check and the published values agree.

notthetup (approved):

PR #443, opened 2026-08-04
Exception handling and hygiene improvements

### LINKED ISSUE #441: ParameterReq does not respond when specifying an unknown parameter enum constant
fjage version: 2.6.0 (unet 7.0.0-rc3)
fjage gateway library: custom

Log file excerpt:
```
Failed to process message: {"id":"84bbe830-6530-492b-ae40-d49493451c78","action":"send","relay":true,"message":{"clazz":"org.arl.fjage.param.ParameterReq","data":{"msgID":"b8cf184d-3db0-411f-8539-17816e625d44","perf":"REQUEST","recipient":"device","requests":[{"param":"org.arl.unet.device.DeviceParam.health"},{"param":"org.arl.unet.device.DeviceParam.storage"},{"param":"org.arl.unet.device.DeviceParam.thermal"},{"param":"org.arl.unet.device.DeviceParam.vendor"},{"param":"org.arl.unet.device.DeviceParam.model"},{"param":"org.arl.unet.device.DeviceParam.serial"},{"param":"org.arl.unet.device.DeviceParam.time"},{"param":"org.arl.unet.device.DeviceParam.upTime"},{"param":"org.arl.unet.device.DeviceParam.voltage"}],"sender":"gateway-746b0aec-cba7-4e19-a01a-bca339e3752b","sentAt":1785812501766}}}
java.lang.IllegalArgumentException: No enum constant org.arl.unet.device.DeviceParam.thermal
Stack trace: ...
org.arl.fjage.remote.EnumTypeAdapter.read(EnumTypeAdapter.java:88)
```

Observed behaviour:
* No response for the `ParameterReq`

Expected behaviour:
* Perhaps a `ParameterRsp` without the invalid/unknown/unsupported parameters?

This makes it hard to write applications that support backwards-compatibility with older versions.

mchitre:
Five fixes with a common flavor: an exception in a peripheral path causes bad behavior in an unrelated core path. One fix per commit, each with a regression test.

### `Container` — listener exceptions break message delivery

- **Problem:** `send()` invoked `MessageListener.onReceive()` with no guard, so a throwing listener lost the message and the exception surfaced inside the *sending* agent's `send()` call, making it look like a bug in whichever agent happened to be sending.
- **Fix:** Catch and log per listener. The listener set also becomes a `CopyOnWriteArraySet`, removing a lock-ordering hazard: listeners were called while holding the `listeners` monitor, with the container monitor acquired only afterwards, so a listener calling a synchronized container method inverted the lock order.
- **Outcome:** A broken listener is now a logged warning rather than a broken container.

### `EnumTypeAdapter` — unknown enum constant kills the whole message (#441)

- **Problem:** `Enum.valueOf()` throws when a parameter enum constant is unknown to this version, aborting deserialization of the entire `JsonMessage`. The request was never delivered, so the sender got no response at all.
- **Fix:** An unknown constant degrades to a `NamedParameter`, exactly as an unloadable enum class already did. `ParameterMessageBehavior` finds no match and drops that entry, so the response carries every valid parameter and omits the unknown one, matching how mis-named `NamedParameter`s already behave.
- **Caveat:** The fully qualified name is retained, so a removed constant cannot be mistaken for an unrelated parameter with the same short name.

### `WebServer` — NPE on a server with no handlers

- **Problem:** Jetty's `getHandlers()` returns null rather than an empty array when nothing is registered, so `hasHandler()`, `hasStatic()` and both `setErrorHandler()` overloads threw `NullPointerException` on an empty server. Reachable whenever a component registers contexts and later removes them.
- **Fix:** Route the three iterating call sites through a helper that normalizes a null collection and a null handler array to an empty array. The `setErrorHandler()` overloads already guarded the former but not the latter.

### `WebServer` — web socket upgrades on bare context paths

- **Problem:** `ContextHandler` defaults `allowNullPathInfo` to false, so a bare context path gets a 302 to the trailing-slash form. Browsers do not follow redirects on a web socket upgrade, so a client connecting to `/shell/ws` rather than `/shell/ws/` silently fails to connect. Our clients all compensate, but the requirement was undocumented.
- **Fix:** Serve the bare path directly for handlers added via `addHandler()`, matching what `addUpload()` already does.
- **Note:** Deliberately not applied to `addStatic()`, where the redirect is wanted.

### `WebServer` — report when the port is already in use

- **Problem:** A failed `start()` was logged vaguely, leaving callers with a server that quietly serves nothing while every dependent component fails later in unrelated ways. The usual cause, another process holding the port, had to be inferred from a stack trace.
- **Fix:** Name bind failures explicitly, so the log says which port is taken. Startup remains non-fatal, so this changes diagnostics only.

### DIFF
--- a/src/main/java/org/arl/fjage/Container.java
+++ b/src/main/java/org/arl/fjage/Container.java
@@ -57,7 +57,7 @@ public class Container {
   protected Method doClone;
   protected boolean autoclone = false;
   protected final Set<AgentID> idle = new HashSet<>();
-  protected final Set<MessageListener> listeners = new HashSet<>();
+  protected final Set<MessageListener> listeners = new CopyOnWriteArraySet<>();
 
   //////////// Interface methods
 
@@ -320,15 +320,15 @@ public class Container {
   /**
    * Adds a listener to the container. A listener is able to snoop on all
    * messages passing through the container, and optionally able to ask the
-   * container to drop a message.
+   * container to drop a message. A listener that throws an exception has the
+   * exception logged and ignored, so that a misbehaving listener cannot
+   * disrupt message delivery.
    *
    * @param listener listener.
    * @return true if successfully added, false otherwise.
    */
   public boolean addListener(MessageListener listener) {
-    synchronized (listeners) {
-      return listeners.add(listener);
-    }
+    return listeners.add(listener);
   }
 
   /**
@@ -338,9 +338,7 @@ public class Container {
    * @return true if successfully removed, false otherwise.
    */
   public boolean removeListener(MessageListener listener) {
-    synchronized (listeners) {
-      return listeners.remove(listener);
-    }
+    return listeners.remove(listener);
   }
 
   /**
@@ -366,9 +364,12 @@ public class Container {
   public boolean send(Message m, boolean relay) {
     if (relay) log.warning("Container does not support relaying");
     if (m.getSentAt() == null) m.setSentAt(platform.currentTimeMillis());
-    synchronized (listeners) {
-      for (MessageListener listener: listeners)
+    for (MessageListener listener: listeners) {
+      try {
         if (listener.onReceive(m)) return true;
+      } catch (Throwable ex) {
+        log.log(Level.WARNING, "MessageListener: "+ex.toString(), ex);
+      }
     }
     AgentID aid = m.getRecipient();
     if (aid == null) return false;
--- a/src/main/java/org/arl/fjage/Container.java
+++ b/src/main/java/org/arl/fjage/Container.java
@@ -367,8 +367,8 @@ public class Container {
     for (MessageListener listener: listeners) {
       try {
         if (listener.onReceive(m)) return true;
-      } catch (Throwable ex) {
-        log.log(Level.WARNING, "MessageListener: "+ex.toString(), ex);
+      } catch (RuntimeException ex) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/arl/fjage/Container.java
```diff
@@ -366,9 +364,12 @@ public boolean send(Message m) {
   public boolean send(Message m, boolean relay) {
     if (relay) log.warning("Container does not support relaying");
     if (m.getSentAt() == null) m.setSentAt(platform.currentTimeMillis());
-    synchronized (listeners) {
-      for (MessageListener listener: listeners)
+    for (MessageListener listener: listeners) {
+      try {
         if (listener.onReceive(m)) return true;
+      } catch (Throwable ex) {
```
notthetup:
```suggestion
      } catch (RuntimeException ex) {
```

Otherwise this swallows fatal JVM conditions such as `OutOfMemoryError`, `StackOverflowError`, and `ThreadDeath`, then continues message delivery in a potentially corrupted process

notthetup (approved):

src/main/java/org/arl/fjage/Container.java
```diff
@@ -366,9 +364,12 @@ public boolean send(Message m) {
...
+        log.log(Level.WARNING, "MessageListener: "+ex.toString(), ex);
```
notthetup:
```suggestion
        log.log(Level.WARNING, "MessageListener: "+ ex, ex);
```

PR #447, opened 2026-08-07
Make fjagepy support Python 3.9

notthetup:
We had lost Python 3.9 support because of 

> TypeError: unsupported operand type(s) for |. 

This fixes that.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -8,8 +8,8 @@ jobs:
     timeout-minutes: 20
     strategy:
       matrix:
-        python-version: [3.12]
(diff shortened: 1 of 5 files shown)


dixslyf (approved):
Good catch. Does CI currently catch this anywhere? It's hard to keep track of which Python features are available in which versions.

Also, if/when we move to Python 3.14, https://peps.python.org/pep-0649/ will supersede `from __future__ import annotations` - just to take note for the future.

notthetup:
>  Does CI currently catch this anywhere?

It doesn't. We can probably run fjage tests on Python 3.9 to be safe.

Issue #445, opened 2026-08-06
Add `DeleteFileReq` and `RenameFileReq` to the shell agent?


notthetup:
## Summary

Context is from a `unet-web` perspective.

The shell agent currently only has `PutFileReq` and `GetFileReq`. While a file can be deleted using a `PutFileReq` (with an offset of 0 and empty contents IIRC), it's a little awkward.

It would be nice to have a dedicated `DeleteFileReq` to delete a file. And maybe even delete multiple files in one request (in `unet-web`, there's a button to delete all previous logs; I believe we currently either spam `PutFileReq`s or send them sequentially - can't quite remember).

It would also be nice to have a `RenameFileReq` to rename a file. Currently, in `unet-web`, we send a Groovy shell command with an `ShellExecReq` to do this, which feels really error-prone because we have to do string interpolation to input the file paths. Also, for some reason, the shell agent takes a long time to respond when doing it this way (about 800ms).

Another one might be a `ZipFilesReq`, but that may be a bit too out of scope. Similar to renaming files, `unet-web` currently uses a `ShellExecReq` to do this.

## Version

v7.0.0

## Platform

Modems

dixslyf:
https://github.com/org-arl/fjage/pull/449 added a `DeleteFileReq`.

Thinking about it more, I'm not sure if we really want a `RenameFileReq`. Because if there is one, then users might expect a `CopyFileReq`, and then other filesystem operations, and it just spirals from there. So I'm thinking that we should just keep those out of fjage.

PR #449, opened 2026-08-12
Add `DeleteFileReq` to the shell agent

dixslyf:
Part of https://github.com/org-arl/fjage/issues/445. Adds a `DeleteFileReq` to the shell agent to delete a file / directory.

Behaviour should be almost identical to what `PutFileReq` did previously for deletion. If the specified path ends in a path separator, then it is expected that the target to delete is a directory, and the shell agent will delete it recursively (now, we also additionally remove the input stream cache entries that fall under that directory). If not, we expect a file and just delete it and its cache entry.

The main difference is that we now also have an additional check: if the specified path ends with a path separator (i.e., the intent is to delete a directory) but is a normal file on disk, then we respond with FAILURE. Similarly, if the specified path doesn't end with a path separator but is a directory on disk, we also respond with FAILURE.

To avoid code duplication, the handler for `DeleteFileReq` and the deletion code path for `PutFileReq` use the same `deletePath()` private method. This does mean the deletion behaviour of `PutFileReq` will change to include the additional check above.

Also, the input stream cache (`isCache`) is now keyed by a normalised form of the path (via a new `cacheKey()` helper) rather than the raw filename string. This is needed so that a nested file's cache entry can be reliably matched against the directory being deleted, regardless of how each path is formatted (e.g., trailing separators, redundant segments). `GetFileReq` handling has also been updated to use `cacheKey()` for consistency.

Also added tests for `DeleteFileReq`.

### DIFF
--- a/gateways/Gateways.md
+++ b/gateways/Gateways.md
@@ -239,11 +239,12 @@ The `clazz` field should be set based on the type of the base64 array is being e
 
 ## Pre-defined Messages
 
-A fjåge Gateway may export pre-defined Message Types for the Messages defined by fjåge. These are :
+A fjåge Gateway may export pre-defined Message Types for the Messages defined by fjåge. These are:
 
 - `org.arl.fjage.shell.ShellExecReq`
 - `org.arl.fjage.shell.GetFileReq`
 - `org.arl.fjage.shell.PutFileReq`
+- `org.arl.fjage.shell.DeleteFileReq`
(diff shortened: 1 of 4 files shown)


mchitre (approved):
We might want to also update:
- `gateways/Gateways.md`
- `fjagepy/Message.py`
- `gateways/js/src/message.js`

with the new message.

dixslyf:
I'll raise separate PRs for the gateways.

PR #453, opened 2026-08-18
Add `registerMessageClass` to fjage.js

notthetup:
This pull request modernizes and simplifies the handling of custom message classes in the `fjage.js` JavaScript gateway package. The key change is the introduction of a new `registerMessage()` API for registering custom message classes, replacing the older and now deprecated `MessageClass` function. All documentation, examples, and tests have been updated to use the new approach, and additional improvements have been made to the message registration and instantiation logic.

**Bug Fixes:**
- Fixed a typo in `PublishServerAgent.groovy` (`tempearture` → `temperature`).


gateways/js/src/fjage.js
```diff
@@ -1,4 +1,4 @@
-import { MessageClass, Message, GenericMessage, ParameterReq, ParameterRsp, PutFileReq, GetFileReq, GetFileRsp, DeleteFileReq, ShellExecReq} from './message.js';
+import { Message, GenericMessage, ParameterReq, ParameterRsp, PutFileReq, GetFileReq, GetFileRsp, DeleteFileReq, ShellExecReq, registerMessageClass} from './message.js';
 import { Gateway, init} from './gateway.js';
```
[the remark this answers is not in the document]
notthetup: That's fine. We'll uprev MAJOR version for this change.

gateways/js/src/message.js
```diff
@@ -1,248 +1,292 @@
...
+  message.sentAt = 0;
+
+  if (inReplyToMsgOrFields instanceof Message) {
+    message.recipient = inReplyToMsgOrFields.sender;
+    message.inReplyTo = inReplyToMsgOrFields.msgID;
+  } else if (inReplyToMsgOrFields && typeof inReplyToMsgOrFields === 'object') {
+    Object.assign(message, inReplyToMsgOrFields);
+  }
+
+  if (message.constructor.name.toLowerCase().endsWith('req') && perf === undefined &&
```
dixslyf: According to Claude, this is broken because in the minified bundle, the constructor name becomes mangled, so `.endsWith('req')` will almost certainly never be `true`.

I don't think it's a good idea to set fields based on the class name anyway. Probably better to do that in the subclass's constructor explicitly, though that means a little more boilerplate.

gateways/js/src/message.js
```diff
@@ -1,248 +1,292 @@
...
-    this.__clazz__ = 'org.arl.fjage.Message';
-    this.msgID = UUID7.generate().toString();
-    this.perf = perf;
-    this.sender = null;
-    this.recipient = inReplyToMsg ? inReplyToMsg.sender : null;
-    this.inReplyTo = inReplyToMsg ? inReplyToMsg.msgID : null;
+   * @param {Message|Object} [inReplyToMsgOrFields] - message to reply to, or fields to assign
+   * @param {Performative} [perf] - performative of the message
+   */
+  constructor(inReplyToMsgOrFields, perf) {
```
dixslyf: I'm not a fan of "overloading" a single argument like this. Would it work to have a separate static method like `responseFor(Message): Message` instead?

gateways/js/src/message.js
```diff
@@ -1,248 +1,292 @@
...
+export class ParameterReq extends Message {
+  /** @type {string|null} */
+  param = null;
+  /** @type {*} */
+  value = null;
+  /** @type {Array<ParameterReq.Entry>} */
+  requests = [];
+  /** @type {number} */
+  index = -1;
+}
```
dixslyf, line 201: The implicit constructor JS generates for subclasses won't correctly set the fields if we do something like `new ParameterReq({ param: "x", value: 42 })`. `param` and `value` will both be `null` (the default values we defined in the class definition).

This is because the implicit constructor does something along these lines:
```javascript
constuctor(...args) {
  super(...args);
  this. param = null;
  this.value = null;
  this.requests = [];
  this.index = -1;
}
```

The subclass's fields are set by the constructor _after_ calling `super()`, so although `initializeMessage()` sets the fields in `Message`'s constructor, the fields specific to the subclass will get overwritten by the defaults.

Not sure what the best way to fix this is. Perhaps `Message` shouldn't try to be too smart and set all the fields.

gateways/js/src/message.js
dixslyf: Even without the name mangling, this won't work for the now deprecated `MessageClass` because we use an anonymous class:

```javascript
const messageClass = class extends parent {   // anonymous class expression
  constructor(fields={}) { super(fields); }
};
```

The class's name will become `messageClass`.

gateways/js/src/message.js
```diff
@@ -1,248 +1,292 @@
...
+
+  if (inReplyToMsgOrFields instanceof Message) {
+    message.recipient = inReplyToMsgOrFields.sender;
+    message.inReplyTo = inReplyToMsgOrFields.msgID;
+  } else if (inReplyToMsgOrFields && typeof inReplyToMsgOrFields === 'object') {
+    Object.assign(message, inReplyToMsgOrFields);
+  }
+
+  if (message.constructor.name.toLowerCase().endsWith('req') && perf === undefined &&
+      !(inReplyToMsgOrFields && typeof inReplyToMsgOrFields === 'object' && 'perf' in inReplyToMsgOrFields)) {
```
dixslyf: This needs an additional check that `inReplyToMsgOrFields` is _not_ an instance of `Message` because if it's a `Message`, it's interpreted as the message to _reply_ to and in that case we shouldn't be looking at that message's `perf`.

gateways/Gateways.md
```diff
@@ -182,11 +182,30 @@ If a gateway's agent name collides with an existing agent known to the master, t
...
+
+## registerMessageClass() :: String name, Class messageClass -> Void
+
+- Registers a message class in a global registry of message classes, so that when a message of this class is received, an instance of this class is created to represent the message.
+- The `messageClass` must extend the base _Message_ class.
+- The `name` is the fully qualified name of the message class, and is used to identify the message class when a message of this class is received.
+- A received fully qualified name is resolved exactly. A received unqualified name is resolved to the most recently registered class with that short name.
+- A registered message class should support construction without arguments. If it does not, only base _Message_ fields are inflated.
+- A received field that is not already declared on a registered message instance is discarded.
+- A Gateway may extend this function to support additional features decorators or annotations for message classes if supported by the language.
```
dixslyf:
```suggestion
- A Gateway may extend this function to support additional features, decorators or annotations for message classes if supported by the language.
```

gateways/js/test/spec/create-spec.cjs
dixslyf, line 1: We should also test against the minified version of the library to make sure we catch issues like https://github.com/org-arl/fjage/pull/453/changes#r3809789208.

dixslyf (changes requested):

gateways/js/src/message.js
notthetup: Yea. I am  not very happy with the design.

gateways/js/src/message.js
notthetup: Fair enough. We might have to give up on this compat with the Java version of the library.

notthetup:
OK. Finally it's in decent shape. Please help review @dixslyf

gateways/js/src/message.js
```diff
@@ -1,248 +1,270 @@
...
-      props[key] = this[key];
+    const data = {};
+    for (const key of Object.keys(this)) {
+      if (!key.startsWith('_')) data[key] = this[key];
+    }
+    // if not a registered class, set __clazz__ to the actual class name
+    let clazz = this.__clazz__;
+    if (this.__clazz__ === 'org.arl.fjage.Message' && !messageClassForName(this.__clazz__)) {
+      clazz = this.constructor.name;
     }
```
dixslyf: This never fires because `Message` is always registered as `org.arl.fjage.Message` (via `registerMessageClass('org.arl.fjage.Message', Message);`), so `messageClassForName(this.__clazz__)` will always be truthy.

In any case, if it did fire, it would contradict with the readme:

> Every subclass you send needs its own registration. An unregistered subclass inherits its parent's `clazz`, and so goes on the wire as its parent:
>
> ```js
> class TxFrameReq extends Message {}          // not registered
> new TxFrameReq().toJSON().clazz              // 'org.arl.fjage.Message'
> ```

gateways/js/test/spec/fjage.spec.js
```diff
@@ -616,43 +650,126 @@ describe('A Message', function () {
...
+    expect(message.value).toBe(1);
+  });
+
+  it('should serialize an unregistered message with its original class name', function () {
+    class UnregisteredMessage extends Message { value = null; }
+    const message = new UnregisteredMessage();
+    message.value = 1;
+    const json = message.toJSON();
+    expect(json.clazz).toBe('UnregisteredMessage');
+    expect(json.data.value).toBe(1);
```
dixslyf, line 772: This test fails because of https://github.com/org-arl/fjage/pull/453/changes#r3900560233.

gateways/js/src/message.js
dixslyf: If we stick to the readme's defined behaviour, we should also add an additional test for a 3-level class hierarchy, where the "leaf" class is unregistered. E.g., `org.arl.fjage.Message` (registered) > `org.arl.fjage.ChildMessage` (**registered**) > `org.arl.fjage.GrandchildMessage` (**unregistered**). `GrandchildMessage` should have `org.arl.fjage.ChildMessage` as its `clazz`.

dixslyf (changes requested):

gateways/js/src/message.js
notthetup: Right. I think correct logic for this scenario would be:

"If `MyMessage` is a subtype of `Message` but hasn't been registered using `Gateway.registerMessage`, then serialize to JSON with `clazz: MyMessage`, set all the properties as a `data` field in the JSON Message.

Not much point here going out as the registered parent.

dixslyf: Sounds reasonable, but `clazz: MyMessage` would be an unqualified class name, right? Just want to make sure that actually works on a protocol level.

docs/jsgw.qmd
```diff
@@ -39,16 +39,15 @@ Here `WebServer.addStatic()` serves the static web resources bundled in the fjå
 
 ## Use the JavaScript module
 
-fjage.js is distributed as ready-to-use bundles for the common JavaScript module systems (ESM, CommonJS and UMD). In a web application built with a bundler, you simply `import { Gateway, MessageClass } from 'fjage'`; in a plain web page, you can either load the UMD bundle with `<script src="fjage.min.js"></script>`, or import the ESM bundle (`dist/esm/fjage.js`) directly from a `<script type="module">`. Examples for each module system are available in the [examples](https://github.com/org-arl/fjage/tree/master/gateways/js/examples) directory.
+fjage.js is distributed as ready-to-use bundles for the common JavaScript module systems (ESM, CommonJS and UMD). In a web application built with a bundler, you simply `import { Gateway, ShellExecReq } from 'fjage'`; in a plain web page, you can either load the UMD bundle with `<script src="fjage.min.js"></script>`, or import the ESM bundle (`dist/esm/fjage.js`) directly from a `<script type="module">`. Examples for each module system are available in the [examples](https://github.com/org-arl/fjage/tree/master/gateways/js/examples) directory.
```
dixslyf: This makes it sound like you _must_ import `ShellExecReq` from `fjage`. I think importing `Gateway` is sufficient.

docs/jsgw.qmd
```diff
@@ -76,14 +75,14 @@ function makeRq(shell) {
...
+This code first opens a gateway through the web socket interface back to the web server that served this JavaScript (in the browser, `new Gateway()` defaults to the hostname and port of the page, with the WebSocket endpoint at `/ws/`; these can be overridden by passing `{ hostname, port, pathname }` options). It then looks for an agent providing the `org.arl.fjage.shell.Services.SHELL` service. If found, it subscribes to messages from that service and calls `makeRq()` to make a command execution request to the agent providing that service. The request is to execute a command `"ps"` and simply log the response to the browser's console.
 
 ## Use from Node.js
 
 The same API works in Node.js, where the gateway connects to the master container's TCP port instead of a web socket:
 
 ``` javascript
-const { Gateway, MessageClass } = require('fjage');
+const { Gateway, ShellExecReq } = require('fjage');
```
dixslyf: The `ShellExecReq` import isn't used in the example.

gateways/js/test/spec/minified-spec.cjs
dixslyf, line 1: This looks suspect. Does the whole test suite actually run with the minified bundle? It looks like it only runs this one test here.

gateways/js/test/run-tests.cjs
```diff
@@ -45,7 +45,7 @@ if (serveBrowserTest) {
   if (runNodeTest) {
     console.log('Running jasmine..');
     let startTime = new Date();
-    await runJasmine('test/spec/', ['fjageSpec.cjs']);
+    await runJasmine('test/spec/', ['fjageSpec.cjs', 'minified-spec.cjs']);
```
dixslyf: `minified-spec.cjs` no longer exists.

dixslyf (approved):

PR #454, opened 2026-08-18
Deprecate `MessageClass` in fjagepy.

notthetup:
Add a warning to `MessageClass` function, indicating that it's deprecated.


gateways/python/fjagepy/Message.py
```diff
@@ -220,6 +221,7 @@ def __init__(self):
         self.__clazz__ = "org.arl.fjage.GenericMessage"
 
 
+@deprecated("Use @message decorator or registerMessageClass instead to register message classes for JSON inflation.")
```
dixslyf: There is no `registerMessageClass`, but there is [`_register_message_class()`](https://github.com/org-arl/fjage/blob/7a71ef091868918eb34e6030c9431b06a28f5d32/gateways/python/fjagepy/Message.py#L53), which is not part of the public API. So we should either remove that mention or make `_register_message_class()` public.

dixslyf (changes requested):

gateways/python/fjagepy/Message.py
```diff
@@ -4,6 +4,7 @@
 import logging
 import keyword
 from typing import Callable, Optional, Any, Dict, TYPE_CHECKING, Union, get_type_hints, overload
+from warnings import deprecated
```
dixslyf: Seems like this doesn't exist on Python 3.9.

An alternative is to pull in [`typing_extensions`](https://pypi.org/project/typing-extensions/) (an official package from Python), which backports some of the newer type features to older Python versions. It also includes `deprecated()`. `typing_extensions` might be useful for other stuff as well.

Or we could just define our own `deprecated()` decorator.

dixslyf (approved):

notthetup:
Moving this change to another branch

PR #455, opened 2026-08-19
Define message registration API for Gateways

notthetup:
This pull request updates the gateway documentation to clarify and modernize how message types are registered and filtered. The changes deprecate the old `MessageClass` mechanism in favor of a new `registerMessage()` API and clarify filtering options for messages.

**Message registration and serialization improvements:**

* Added a new `registerMessage()` API for registering `Message` subclasses with serialization/deserialization, specifying how fully qualified names are used. Also notes support for language-specific registration mechanisms like annotations or decorators.
* Marked the old `MessageClass` mechanism as deprecated, recommending the new registration API for custom message types.
* Updated instructions for creating custom message types to recommend subclassing `Message` and using the new registration approach for correct gateway serialization.

**Message filtering clarification:**

* Clarified that message filtering may use a `Message` subclass or language-equivalent type, rather than the previous reference to `MessageClass class`.


dixslyf (approved):
BTW, just curious, but does the message registry have to be global? AFAIK, our gateway libraries always store it in some global object, which feels iffy. Why not e.g. store the registry in the `Gateway` object itself (where each `Gateway` object has its own message registry) and provide a `registerMessage()` method?

notthetup:
> BTW, just curious, but does the message registry have to be global? AFAIK, our gateway libraries always store it in some global object, which feels iffy. Why not e.g. store the registry in the Gateway object itself (where each Gateway object has its own message registry) and provide a registerMessage() method?

It's possible that a single Python/JavaScript app has multiple Gateway instances that connect to different fjage MasterContainers. Would that require the custom messages to be registered for EVERY Gateway instance? That feels a bit awkward.

Another option to have a static API on the Gateway class. That might not be available in all languages. That's also similar to Java Gateway API.

dixslyf:
> Would that require the custom messages to be registered for EVERY Gateway instance? That feels a bit awkward. 

Yeah, that was what I was thinking, because one `Gateway` might care about a specific type of message while another might not. But I agree, it's a little awkward and likely not even a common use case (and registering globally already satisfies that use case anyway if you just filter).

---

On static methods, perhaps we could update the spec to say that libraries should prefer making it a static method if the language supports it, otherwise fall back to a plain function?

dixslyf (approved):

mchitre (commented):
There's additional complexity we have to handle, with message sub-classing. How do we specify that in the spec?

notthetup:
> There's additional complexity we have to handle, with message sub-classing. How do we specify that in the spec?

Is this capturing the information that `TxFrameReq` -> `DatagramReq` -> `Message` ? Or?

mchitre:
> > There's additional complexity we have to handle, with message sub-classing. How do we specify that in the spec?
> 
> Is this capturing the information that `TxFrameReq` -> `DatagramReq` -> `Message` ? Or?

Yes. Things like `RemoteMessageNtf <: DatagramNtf`, `TxFrameReq <: DatagramReq`, etc. This hierarchy is often important to represent in the type system, otherwise things like `instanceof` give wrong answers.

notthetup:
@mchitre All 3, the updated Python implementation (https://github.com/org-arl/fjage/pull/456), the new JS implementation (https://github.com/org-arl/fjage/pull/453) and https://github.com/org-arl/UnetSockets.jl/blob/master/src/Messages.jl all allow using of the type system to define the parent classes. So...

```js
class RemoteMessageNtf extends DatagramNtf {
    ...
}

Gateway.registerMessage('org.arl.unet.RemoteMessageNtf', RemoteMessageNtf);
```


```py


class RemoteMessageNtf(DatagramNtf):
    pass

Gateway.registerMessage("org.arl.unet.RemoteMessageNtf", RemoteMessageNtf)

# OR

@message("org.arl.unet.RemoteMessageNtf")
class RemoteMessageNtf(DatagramNtf):
    pass

```

and

```jl
Fjage.@message "org.arl.unet.remote.RemoteMessageReq" :REQUEST struct RemoteMessageReq <: DatagramReq
...
end
```


I've captured that a Gateway in a language that doesn't support this, should extend the API to allow defining parent classes.

mchitre (commented):
1. "Most recently registered wins" is order-dependent global state, and load order across classes or packages may not be deterministic in some languages. Better to be silent on expected behavior, and leave it to an implementation to decide.

2. Java/Groovy always sends fully qualified names. Bare names only appear when a non-Java gateway serializes a class that was registered under an unqualified name. That means the spec is adding receive-side complexity to accommodate sloppy send-side registration. Unqualified class names also won't inflate in Java if the class is in a package. We should perhaps insist on gateways registering with qualified names. Lookups can be unqualified, but registrations should always use fully qualified class names.

mchitre (approved):
1. Collision handling is claimed but not specified. The PR description says the API specifies "how name collisions are handled", but we no longer do that, so we should remove that claim in the description.
2. Fallback should preserve the original clazz. The GenericMessage fallback wording only requires a `Map<String,Object>` of fields. `Fjage.jl` deliberately keeps the class name in the type parameter so `classname(msg)` still returns e.g. `org.arl.unet.FooNtf` and re-serialization round-trips correctly. I'd suggest the spec explicitly require the fallback to retain the original `clazz` — otherwise a JS/Python implementation could lose that.

notthetup:
> Collision handling is claimed but not specified. The PR description says the API specifies "how name collisions are handled", but we no longer do that, so we should remove that claim in the description.

Updated.

> Fallback should preserve the original clazz. The GenericMessage fallback wording only requires a Map<String,Object> of fields. Fjage.jl deliberately keeps the class name in the type parameter so classname(msg) still returns e.g. org.arl.unet.FooNtf and re-serialization round-trips correctly. I'd suggest the spec explicitly require the fallback to retain the original clazz — otherwise a JS/Python implementation could lose that.

Good catch. I'll add that.

PR #456, opened 2026-08-19
Deprecate MessageClass factory and add `register_message` to fjagepy

notthetup:
This pull request modernizes and clarifies the way custom message classes are registered and handled in the Python gateway, deprecating the old `MessageClass` helper in favor of a new, more robust registration mechanism. It introduces a new `Gateway.registerMessage()` static method and improves the `@message` decorator, ensuring better interoperability and clearer semantics for message serialization and deserialization. The documentation and tests are updated accordingly, and several internal improvements are made to the message registration and inflation logic.


gateways/python/fjagepy/Message.py
```diff
@@ -50,9 +51,23 @@ def _message_clazz(class_: type["Message"]) -> str:
         return clazz_name
     return "org.arl.fjage.Message"
 
-def _register_message_class(class_: type["Message"], fqcn: Optional[str] = None) -> type["Message"]:
-    if not issubclass(class_, Message):
-        raise TypeError('@message can only be used with Message subclasses')
+def registerMessage(qualified_name: str, message_class: type["Message"]) -> type["Message"]:
```
dixslyf: Python uses snake case, so we should rename this to `register_message`.

notthetup: Good point. I was trying to keep compat with the updated Gateway spec 

https://github.com/org-arl/fjage/pull/455. 

Generally we try to adhere to the original Java naming schemes for anything that's designed to be copy of the Java API. But there is no similar API on the Java Gateway, so I guess we can keep the Java style naming convention for all the Gateways.

dixslyf: Right, then actually let's keep it as `registerMessage()`. I just noticed that we've already been using camel case for fjagepy, so best to stick to that rather than having some functions in camel case vs. snake case.

notthetup:
OK. Finally it's in good shape. Please help review @dixslyf

gateways/python/fjagepy/Message.py
```diff
@@ -152,18 +153,20 @@ def from_json(cls, json_obj: Dict[str, Any], owner: Optional["Gateway"] = None)
...
         clazz_name = qclazz.split(".")[-1]
 
-        rv_cls = _MESSAGE_REGISTRY.get(qclazz) or _MESSAGE_REGISTRY.get(clazz_name)
+        registered_cls = _MESSAGE_REGISTRY.get(qclazz)
+        rv_cls = registered_cls
 
         if rv_cls is None:
-            module = sys.modules[__name__]
-            rv_cls = getattr(module, clazz_name, None)
+            reflected_cls = globals().get(clazz_name)
```
dixslyf: `globals()` only returns the globals in the _current_ module (in this case, `Message.py`), so this is too narrow and not going to be useful.

You can probably iterate over the modules in `sys.modules`, but that's an expensive operation we don't want to keep paying for.

Since the expectation is to register message classes using `Gateway.registerMessage()` or `@message`, perhaps we should drop the reflection for the Python gateway?

gateways/python/fjagepy/Gateway.py
```diff
@@ -44,6 +44,13 @@ class Gateway:
     NON_BLOCKING = 0
     BLOCKING = -1
 
+    @staticmethod
+    def registerMessage(class_name: str, message_class: type[Message]) -> type[Message]:
```
dixslyf: On naming conventions again... I just noticed that we generally provide both camel case and snake case variants of the same method.

In some cases, we've also deprecated the camel case counterpart and forwarded the reader to the snake case version (e.g., `isConnected()` vs. `is_connected()`).

Perhaps a separate discussion to have on naming conventions (though my preference is to use the conventions of the language), but what should we do for `registerMessage()`?

gateways/python/fjagepy/Message.py
```diff
@@ -252,16 +264,17 @@ def message(arg: str) -> Callable[[type[Message]], type[Message]]: ...
...
     """
 
     def decorate(class_: type["Message"], fqcn: Optional[str] = None) -> type["Message"]:
         if not issubclass(class_, Message):
             raise TypeError('@message can only be used with Message subclasses')
 
-        clazz_name = fqcn or class_.__dict__.get('__clazz__') or class_.__name__
-        class_.__clazz__ = clazz_name
-        return _register_message_class(class_, clazz_name)
+        return register_message(fqcn or class_.__dict__.get('__clazz__') or class_.__name__, class_)
```
dixslyf: IIRC, the gateway spec requires registration to be under a fully qualified class name, in which case I don't think it's a good idea to fall back to `class_.__name__` because that's unqualified.

dixslyf: Perhaps we should just remove the use of the `@message` decorator without an argument?

i.e., disallow this:

```python
@message
class CustomReq(Message):
    pass
```

and force callers to do this:

```python
@message('org.example.CustomReq')
class CustomReq(Message):
    pass
```

dixslyf (changes requested):

gateways/python/fjagepy/Message.py
notthetup: On way to deal with this could be to "iterate over the modules in `sys.modules`" and then if we find something, register it, so that we don't pay for the iteration cost everytime we see the same message.

gateways/python/fjagepy/Gateway.py
notthetup: Agreed it's not consistent yet. Since the last rewrite I have tried to keep to the naming conventions as close to what's defined in the Gateway.md spec. The old `is_connected` was deprecated and left there. But maybe we take this opportunity to force all the external facing APIs to follow Gateway spec and remove the older versions.

gateways/python/fjagepy/Message.py
notthetup: Yea. That's a good call. I'll fix that.

gateways/python/fjagepy/Gateway.py
dixslyf: I believe it was the Java-style name that was deprecated (`isConnected()`), not the Python-style one (`is_connected()`), which makes it awkward because if we want to stick with the Java naming conventions, then we'll have to "swap" the deprecation.

notthetup: Ah yes. Now I remember the thinking. 

We should allow Gateways to mangle method names so they're idiomatic to the Gateway's language. The thinking behind this is that some languages may have strict restrictions on naming. So instead of trying to follow Java conventions being neither consistent with Java nor idiomatic in the language, we're better off naming the API naturally in the language of the Gateway.

We should add this to Gateway.md. (will raise a seperate PR for this)

And I'll update `registerMessage` to `register_message`.

gateways/python/fjagepy/Message.py
dixslyf: I guess that works. Feels awkward though.

We should document this reflection-based auto-registration mechanism in the readme if we implement it.

notthetup: I agree it feels awkward, let's skip that for now and document that decision.

gateways/python/fjagepy/Message.py
```diff
@@ -241,33 +244,21 @@ def __init__(self, **kwargs):
...
-        if not issubclass(class_, Message):
-            raise TypeError('@message can only be used with Message subclasses')
-
-        clazz_name = fqcn or class_.__dict__.get('__clazz__') or class_.__name__
-        class_.__clazz__ = clazz_name
-        return _register_message_class(class_, clazz_name)
-
-    if isinstance(arg, type):
-        return decorate(arg)
+    if not isinstance(class_name, str) or '.' not in class_name or any(not part for part in class_name.split('.')):
```
dixslyf: See https://github.com/org-arl/fjage/pull/456/changes#r3910205436

gateways/python/fjagepy/Gateway.py
```diff
@@ -44,6 +44,13 @@ class Gateway:
     NON_BLOCKING = 0
     BLOCKING = -1
 
+    @staticmethod
+    def register_message(class_name: str, message_class: type[Message]) -> type[Message]:
+        """Register a Message subclass under a fully qualified wire name."""
+        if not isinstance(class_name, str) or '.' not in class_name or any(not part for part in class_name.split('.')):
```
dixslyf: The code for this check is repeated in the `message()` decorator (https://github.com/org-arl/fjage/pull/456/changes#r3910203899). Extract into a utility function?

dixslyf (approved):

